PR #17, opened 2025-03-05
fix: fullscreen on ios

### LINKED ISSUE #2: [mobile] fullscreen doesn't do anything

ImLunaHey:
iOS devices cannot use fullscreen most HTML elements apart from the video element.

so with this we detect if they're on iOS and then only try the video. if we're on any other device we try the div and then fallback to the video.

we can also drop the `props.setFullscreen` calls in the `setFullscreen` function as we have a listener further down that handles that.

closes https://github.com/streamplace/streamplace/issues/2

### DIFF
--- a/js/app/components/player/video.tsx
+++ b/js/app/components/player/video.tsx
@@ -2,10 +2,10 @@ import Hls from "hls.js";
 import {
   ForwardedRef,
   forwardRef,
-  RefObject,
   useEffect,
   useState,
   useCallback,
+  MutableRefObject,
 } from "react";
 import { View } from "tamagui";
 import {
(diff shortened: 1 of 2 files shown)


js/app/components/player/video.tsx
```diff
@@ -279,6 +279,7 @@ export function WebcamIngestPlayer(
   const handleRef = useCallback((node: HTMLVideoElement | null) => {
     if (node) {
       setVideoElement(node);
+      props.videoRef.current = node;
```
ImLunaHey, line 282: this was the main thing causing issues. 😅 

i tried and tried only to realise the `videoRef` was null the whole time lol

iameli:
Haven't been able to get this working sadly :(

ImLunaHey:
@iameli let me take another stab at it this week/weekend.

PR #64, opened 2025-04-01
fix native fullscreen for webRTC

### LINKED ISSUE #2: [mobile] fullscreen doesn't do anything

10xCrazyHorse:
on native it hides the buttons on fullscreen so this mimics the fullscreen behavior and adds the controls overlay for webRTC only. HLS uses native fullscreen method

closes #2

### DIFF
--- a/js/app/components/settings/settings.tsx
+++ b/js/app/components/settings/settings.tsx
@@ -2,22 +2,43 @@ import {
   selectTelemetry,
   setURL,
   telemetryOpt,
+  selectStreamplace,
 } from "features/streamplace/streamplaceSlice";
 import useStreamplaceNode from "hooks/useStreamplaceNode";
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { useAppDispatch, useAppSelector } from "store/hooks";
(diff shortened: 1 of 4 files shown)


js/app/components/settings/settings.tsx
```diff
@@ -2,22 +2,43 @@ import {
...
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { useAppDispatch, useAppSelector } from "store/hooks";
-import { Button, Form, H3, Input, View, XStack, Label } from "tamagui";
+import { Button, Form, H3, Input, View, XStack, Label, Text } from "tamagui";
 import { Updates } from "./updates";
 import { Switch } from "react-native";
 
+const DEFAULT_PROD_URL = "https://stream.place";
+const DEFAULT_DEV_URL = "http://10.0.1.5:38080";
```
iameli, line 15: These are defined in https://github.com/streamplace/streamplace/blob/next/js/app/.env.production and https://github.com/streamplace/streamplace/blob/next/js/app/.env.development and would prefer to keep them there

If you need to override them locally at build-time you can create an `.env.developement.local` file and it should get picked up

iameli:
merged upstream

PR #75, opened 2025-04-15
add toggle chat panel

10xCrazyHorse:


iameli:
Not deleted on purpose, just deleted that branch! Plz rebase against `next` and re-open

PR #80, opened 2025-04-16
add volume slider

### LINKED ISSUE #56: player: volume slider (web only)

10xCrazyHorse:
adds a vertical popover for volume control fixes #56 but it for both mobile and web


iameli:
I don't think we want a volume slider for mobile - I know there's probably different opinions on this but I always expect to adjust volume for mobile with the hardware buttons. Will review for web though!

iameli (changes requested):
LGTM on web, will merge if we take it out of mobile (or people otherwise convince me that users expect a volume slider on mobile)

10xCrazyHorse:
disabled for mobile

iameli (approved):
LGTM!

PR #81, opened 2025-04-16
add toggle for using custom node

10xCrazyHorse:
minor ui changes on settings page


iameli (changes requested):
Nitpick with the interface here: http://foo.bar is definitely a custom node. So why is the slider unchecked? Shouldn't turning it off return you to `https://stream.place` or whatever the default is?

image

PR #87, opened 2025-04-18
chat: add replies & autocomplete

### LINKED ISSUE #25: chat: show your chats instantly, don't wait for relay

10xCrazyHorse:
fixes #25 #26 #88 #94


js/app/package.json
```diff
@@ -32,6 +32,7 @@
     "@atproto/oauth-client": "^0.3.1",
     "@bacons/text-decoder": "^0.0.0",
     "@rainbow-me/rainbowkit": "2.2.0",
+    "@react-native-async-storage/async-storage": "^2.1.2",
```
iameli: Can't do this - would need a new native build. Use the storage singleton class instead (which uses expo-sqlite under the hood on native) https://github.com/streamplace/streamplace/blob/next/js/app/storage/storage.tsx

Usage looks like this: https://github.com/streamplace/streamplace/blob/39408841320c40bca14508b0608dd6090fe6ca91/js/app/features/bluesky/blueskySlice.tsx#L195-L199

iameli (changes requested):

js/app/package.json
10xCrazyHorse: replaced. builds are failing because replaced the yarn.lock with the one from next

js/app/features/player/playerSlice.tsx
```diff
@@ -99,29 +99,86 @@ const reduceChat = (
...
+        (serverMsg) =>
+          serverMsg.record.text === localText &&
+          !serverMsg.uri.startsWith("local-"),
+      );
+
+      if (matchingServerMsg) {
+        // Found a matching server message, mark this local message for removal
+        localMessageUris.add(msg.uri);
+
+        // [HACK] This ensures reply information is preserved even if the server message doesn't have it
```
iameli: this can go away now 🎉

js/app/components/chat/chat-events.ts
```diff
@@ -0,0 +1,13 @@
+import { EventEmitter } from "events";
+import { MessageViewHydrated } from "features/player/playerSlice";
+
+// Create a global event emitter for cross-platform event handling
+const globalEventEmitter = new EventEmitter();
```
iameli: Could this file be handled as a Redux action instead? And new state that shows the reply prompt?

js/app/components/chat/chat.tsx
```diff
@@ -9,17 +8,110 @@ import {
...
+  View,
+} from "tamagui";
 import { RichText } from "@atproto/api";
+import { chatEvents } from "./chat-events";
 import { ReactElement } from "react";
+import Storage from "storage";
+
+const REPLY_INFO_STORAGE_KEY = "streamplace_reply_info";
+
+async function storeReplyInfoAsync(
```
iameli: We can lose this stuff now 'cause replies are on the relay

iameli: And we can just keep "local-first" message rendering in memory in the Redux state, I don't care if it briefly goes away if you refresh

js/app/components/chat/chat.tsx
```diff
@@ -294,36 +655,3 @@ const RichTextMessage = ({
 
   return <Text>{parts}</Text>;
 };
-
-const ChatMessageText = ({ message }: { message: MessageViewHydrated }) => {
```
iameli, line 298: Intentionally moved out of this component?

iameli (changes requested):
Amazing start, changes per discussion

iameli (changes requested):
There's a weird grey line between some posts now:

image

Also the spacing got smaller compared to prod:

image

Compared with:

image

And I think I liked where it was in the old one. Close tho!

Do you want to just add https://github.com/streamplace/streamplace/pull/98 into this PR? They don't merge cleanly with each other so we may as well I think

PR #91, opened 2025-04-23
media: fix gstreamer leaks for everybody forever

iameli:


iameli:
Not sure why CI is failing here, it's looking okay on git.stream.place. Merging but will investigate

PR #120, opened 2025-05-02
docs: add and embed within final build

### LINKED ISSUE #111: docs: set up basic docs framework

espeon:
should definitely close #111, but unsure about #116 wrt implementation. right now we just move the docs dist folder to the main react native app's dist directory, but a better/more resilient option might be creating a docs snapshot fs and serving that way?


iameli (changes requested):
This looks awesome! Definitely approve. List of minor things to do before this gets merged in:

* Let's get the default port for the docs dev server set similarly to the other ones - Streamplace is `38080` and the Expo dev server is `38081`, so perhaps `38082`?
* Weird one... at least on my Mac the docs seem to be listening on ipv6 only? `http://127.0.0.1:4321/docs` doesn't work, but `http://localhost:4321` does. The main dev app is _required_ to use 127.0.0.1 for OAuth purposes, so it'd be nice to make sure we can use that for everything
* I think I'd like the sidebar visible at all times, currently it's not on the splash page. Maybe that'll change once there's too many docs to fit in one sidebar but for now it's nice to get a picture all at once

Past that, if we remove the boilerplate I'm totally willing to ship this whenever! Even just the lexicon docs is better than no docs

EDIT: Oh, we should also totally get the lexicon stuff auto-generating when we run `make lexicons`, that's where all of the other codegen happens

js/docs/src/content/docs/guides/streamplace-dev-setup.md
```diff
@@ -0,0 +1,32 @@
...
+get started:
+
+## Prerequisites
+
+- [Node.js](https://nodejs.org/)
+- [Yarn](https://yarnpkg.com/)
+  - A way to install it is with `pnpm/npm install -g yarn`. No corepack in this
+    repo _just_ yet.
+- Go (version 1.23)
+  ([We're blocked on going to Go 1.24](https://github.com/NordSecurity/uniffi-bindgen-go/issues/66).)
```
iameli: No longer!
```suggestion
- Go (version 1.24)
```

js/docs/src/content/docs/guides/streamplace-dev-setup.md
```diff
@@ -0,0 +1,32 @@
...
+  - If you use `mise`, you can install latest Go 1.23 with
+    `mise install go@prefix:1.23`
```
iameli:
```suggestion
  - If you use `mise`, you can install latest Go 1.24 with
    `mise install go@prefix:1.24`
```

js/docs/src/content/docs/guides/streamplace-dev-setup.md
```diff
@@ -0,0 +1,32 @@
...
+---
+
+So, you've decided to contribute to Streamplace development. Here's how you can
+get started:
+
+## Prerequisites
+
+- [Node.js](https://nodejs.org/)
+- [Yarn](https://yarnpkg.com/)
+  - A way to install it is with `pnpm/npm install -g yarn`. No corepack in this
```
iameli: Huh, I totally thought we did. Maybe I just had corepack on one of the dev computers and it installed it anyway? PRs welcome!

espeon:
will work on autogeneration either tonight or tomorrow 👍🏼 i don't think it will be particularly hard luckily

js/docs/src/content/docs/guides/streamplace-dev-setup.md
espeon: actually looking at the package.json again I think it is enabled and that my node install from pnpm just didn't include it haha

espeon: changed wording to 'if corepack isn't enabled in your node install', hopefully that works for now

espeon:
(will merge main in? hopefully that fixes the single failing job)

iameli (approved):
Nice! Great start that we can build on

PR #139, opened 2025-05-07
make sure ../atproto is present

kjloveless:
it wasn't immediately clear to me that a local copy of the atproto repo (at least the lexicons) was needed before being able to run make lexgen-types

this does result in the repo being cloned twice during make go-lexicons though


iameli:
oh hey good point, forgot about this entirely. we should probably make it a Meson dependency so it can be versioned though - creating a wrapfile will ensure everyone's got the same version checked out https://github.com/streamplace/streamplace/tree/next/subprojects

iameli:
or an npm dependency, I suppose. are the atproto lexicons published to npm? feels like they should be

kjloveless:
nothing on npm is jumping out at me..

i can take a crack at making it a meson dependency if you're okay with that?

iameli:
by all means! shouldn't be too hard - i forget if you can use meson to clone arbitrary git repos that don't contain a `meson.build` file. but if you can't, it'd be really easy to use a patchfile to insert a no-op meson.build file, take a look at this for an example: https://github.com/DDVTECH/mistserver/blob/f0f61ba3115b93edae58b5a8d1c3d9fa1f29c791/subprojects/mbedtls.wrap#L5

once you make the file you'll need to reference it in `meson.build` so it gets cloned but you won't actually need to do anything with it... so something like this but with even less going on https://github.com/streamplace/streamplace/blob/93180191eae6e30baf69d56f702764c6caf76af5/meson.build#L187-L190

kjloveless:
@iameli this should work, let me know what you think! :D

kjloveless:
ugh... 

i'm curious if the version of meson i used was moving the meson.build from subprojects/packagefiles/atproto to subprojects/atproto

i think that's the only thing that needs to get changed

kjloveless:
... tfw you forgot to even add a directory to the repo

iameli:
@kjloveless Nice! I'm gonna add a CI job to make sure `make lexicons` can run successfully but this is otherwise LGTM

iameli (approved):
LGTM, will merge when CI passes

Makefile
```diff
@@ -132,7 +133,12 @@ lexgen-types:
 		--prefix place.stream \
 		--build-file util/lexgen-types.json \
 		lexicons/place/stream \
-		./subprojects/atproto/lexicons
+		../atproto/lexicons
```
kjloveless: clobbered the updated path to atproto/lexicons

kjloveless (commented):
whoopsie :)

kjloveless:
@iameli i'm feeling lucky this time

iameli:
no worries, CI doing its job!

kjloveless:
￼￼￼

PR #151, opened 2025-05-09
Feature/ratelimit middleware

bluwumeaway:
uses `/x/time/rate` which was already a dependency for the implementation of the ratelimit. there are separate rate limiters for both the messages inside of the websocket vs "normal" api endpoints. From my testing the current limits seem to work/not cause issues with the site in development.


pkg/api/api.go
```diff
@@ -669,6 +682,28 @@ func (a *StreamplaceAPI) HandleLivestream(ctx context.Context) httprouter.Handle
 	}
 }
 
+func (a *StreamplaceAPI) RateLimitMiddleware(ctx context.Context) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			// XXX: check if x-forwarded-for
```
iameli: not a concern for streamplace in production, we host our own HTTPS, but not impossible someone else could configure a proxy for sure

pkg/api/websocket.go
```diff
@@ -203,6 +210,27 @@ func (a *StreamplaceAPI) HandleWebsocket(ctx context.Context) httprouter.Handle
...
+			// wait for rate limit delay if there is one
+			delay := r.Delay()
+			if delay > 0 {
+				select {
+				case <-time.After(delay):
+				case <-ctx.Done():
+					return
+				}
+			}
+
```
iameli: Not worried about this case - this is user --> streamplace messages, and zero of those exist right now - but we should have an IP-specific check for "number of websockets currently open" to handle that one person that's spamming us

pkg/api/api.go
```diff
@@ -669,6 +682,28 @@ func (a *StreamplaceAPI) HandleLivestream(ctx context.Context) httprouter.Handle
...
+			ip, _, err := net.SplitHostPort(req.RemoteAddr)
+			if err != nil {
+				ip = req.RemoteAddr
+			}
+
+			limiter := a.getLimiter(ip)
+
+			if !limiter.Allow() {
+				log.Warn(ctx, "rate limit exceeded", "ip", ip, "path", req.URL.Path)
```
iameli, line 699: note to self: add to grafana

iameli (changes requested):
Really great start! The one missing piece is "open websockets per IP address" which I don't think will be handled by the "requests per second" handler; we'd need something that counts up when a websocket upgrades and counts down when the websocket closes

bluwumeaway:
I removed the to server ratelimiting, and added a conn tracker, but I think it should be configured via the `config` package instead of hardcoded to 5. but i dont know what it would be called in that and it is kind of confusing

iameli (approved):
LGTM, will merge when CI passes. TY!

iameli:
> I removed the to server ratelimiting, and added a conn tracker, but I think it should be configured via the `config` package instead of hardcoded to 5. but i dont know what it would be called in that and it is kind of confusing

I'm cool shipping this for now but yeah a setting would prolly be good

Issue #14, opened 2025-03-05
chore: add more Go linting


iameli:
In particular, I'm interested in the kind of Go linting that will tell us when we're not checking if `err` is nil on the return type — that would have caught several problems for us in the past. I think this is [golangci-lint](https://golangci-lint.run/) maybe?

iameli:
this is done, ty @bluwumeaway, not sure why i can't assign this ticket to you because i just successfully assigned https://github.com/streamplace/streamplace/issues/138 to you 🤷

PR #152, opened 2025-05-10
add gofmt check to see if files could be formatted

bluwumeaway:
should close issue #14 could also automatticaly just apply formatting and run a `go fmt ./...` but i'm not a huge fan of changes being made automatically without seeing what they are.


iameli:
LGTM, thank you! I'm gonna leave #14 open; I added a little more description there for what I was actually looking for. But this is an improvement!

Issue #129, opened 2025-05-04
bug: going live fails when thumbnail is too large sometimes


iameli:
Results in this error:
Error creating livestream
Cannot read properties of undefined (reading 'uri')

<img width="356" alt="Image" src="https://github.com/user-attachments/assets/bb0a52d3-3451-41e1-a3a2-b85a7d6c0fcd" />

Need to instead take a snapshot of the stream client-side, compress, upload that

PR #153, opened 2025-05-10
app: custom live thumbnail + client-side thumbnail generation

### LINKED ISSUE #147: golive: Allow for custom golive image upload
This can be entirely client-side — pick an image to "upload" and then turn around and pass that to `uploadBlob` when we're creating the golive Bluesky post

Bonus points if you let the user pick from the last few!

10xCrazyHorse:
fixes #147 #129

### DIFF
--- a/js/app/features/bluesky/blueskySlice.tsx
+++ b/js/app/features/bluesky/blueskySlice.tsx
@@ -26,6 +26,7 @@ import { privateKeyToAccount } from "viem/accounts";
 import { createAppSlice } from "../../hooks/createSlice";
 import { BlueskyState } from "./blueskyTypes";
 import createOAuthClient from "./oauthClient";
+import { captureVideoFrame, findVideoElement } from "utils/videoCapture";
 
 const initialState: BlueskyState = {
   status: "start",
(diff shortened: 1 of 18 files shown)


js/app/utils/videoCapture.ts
```diff
@@ -0,0 +1,69 @@
+/**
+ * Utility functions for capturing and compressing video frames
+ */
+
+/**
+ * Captures a frame from a video element and returns it as a compressed PNG blob
```
iameli:
```suggestion
 * Captures a frame from a video element and returns it as a compressed JPEG blob
```

js/app/utils/videoCapture.ts
```diff
@@ -0,0 +1,69 @@
...
+      );
+    } catch (error) {
+      reject(error);
+    }
+  });
+};
+
+export const findVideoElement = (): HTMLVideoElement | null => {
+  return document.querySelector("video");
+};
```
iameli: We can do better than this (and there might be more than one on the page someday), I'll get it exposed from the player

iameli (changes requested):
Kinda out of vertical space on the layout here, we're going to need to move some things side-by-side
image

iameli:
@10xCrazyHorse You can use this to get at the correct <video> element without a querySelector https://github.com/streamplace/streamplace/pull/153/files#diff-93d76faf77dd3233bec36edfd5e80c216201b4a34b81b8e3be2442b028bfdeb2R30

js/app/features/bluesky/blueskySlice.tsx
```diff
@@ -58,20 +58,40 @@ const uploadThumbnail = async (
...
-    throw new Error(`failed to fetch thumbnail (http ${thumbnailRes.status})`);
+  if (customThumbnail) {
+    try {
+      const thumbnail = await pdsAgent.uploadBlob(customThumbnail);
+      if (thumbnail.success) {
+        console.log("Successfully uploaded custom thumbnail");
+        return thumbnail.data.blob;
+      }
+    } catch (e) {
+      console.error("Error uploading custom thumbnail:", e);
```
iameli: Probably want to just `throw` this — that'll make it pop up to the user instead of silently falling back to the server thumbnail which they weren't looking for

iameli (changes requested):
Still a little harsh on the vertical space:

image

Other than that let's wait until https://github.com/streamplace/streamplace/pull/158 passes and rebase off of that to make sure mobile build is ok

iameli:
Just waiting on #158 to land

iameli:
@10xCrazyHorse Yeah, Android build failures. You can look into it but I'm also totally happy to ship this web-only for now if you wanna just take out the `expo-image-picker` stuff

iameli (approved):

iameli:
Oops, this doesn't work with server-side OAuth - my fault

PR #164, opened 2025-05-13
fix: sidebar slice app error on sim

espeon:
on first launch, the app errors out due to some sidebar-related items not being available. haven't encountered this on a real device just yet but it's possible it may occur? after the first launch, it works fine.


espeon:
tested again, doesn't work so not ready to merge yet 🙇🏼

iameli:
@espeon converted to draft

espeon:
still can not pin down the issue so closing for now

Issue #82, opened 2025-04-16
Stream Uptime indicator


PSingletary:
When a Viewer joins the stream they shoudl be able to quickly identify how long the stream has been live.
When a Streamer/broadcaster is streaming they should know how long they have been streaming for

The implementions of an uptime indicator also needs to take desktop/mobile views in consideration

- so maybe it's just an overlay that appears on tap/cursor hover?
- maybe it's just a # next to the view count?

PR #165, opened 2025-05-13
adds a counting timer to the livestream info bar?\

bluwumeaway:
should close issue #82

### DIFF
--- a/js/app/components/timer.tsx
+++ b/js/app/components/timer.tsx
@@ -32,7 +32,11 @@ export default function Timer({ start }: { start: string | Date }) {
       paddingHorizontal="$2"
       paddingVertical="$1"
     >
-      <Text fontFamily="$mono" textShadowOffset={{ width: -1, height: 1 }} textShadowRadius={3}>
+      <Text
+        fontFamily="$mono"
+        textShadowOffset={{ width: -1, height: 1 }}
+        textShadowRadius={3}
+      >
         {formatTime(elapsedTime)}
       </Text>
     </View>
--- a/js/app/components/timer.tsx
+++ b/js/app/components/timer.tsx
@@ -11,7 +11,7 @@ export default function Timer({ start }: { start: string | Date }) {
       const now = new Date();
       const elapsed = Math.floor((now.getTime() - startDate.getTime()) / 1000);
       setElapsedTime(elapsed);
-    }, 1000);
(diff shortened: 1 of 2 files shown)


js/app/components/timer.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <View
+      justifyContent="center"
+      flexDirection="row"
+      alignItems="center"
+      paddingHorizontal="$2"
+      paddingVertical="$1"
+    >
+      <Text textShadowOffset={{ width: -1, height: 1 }} textShadowRadius={3}>
```
iameli: Let's do a fixed-width font so it doesn't wiggle around as it counts:
```suggestion
      <Text fontFamily="$mono" textShadowOffset={{ width: -1, height: 1 }} textShadowRadius={3}>
```

js/app/components/timer.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const [elapsedTime, setElapsedTime] = useState(0);
+
+  useEffect(() => {
+    const startDate = typeof start === "string" ? new Date(start) : start;
+
+    const interval = setInterval(() => {
+      const now = new Date();
+      const elapsed = Math.floor((now.getTime() - startDate.getTime()) / 1000);
+      setElapsedTime(elapsed);
+    }, 1000);
```
iameli: If it's at 1 second it'll occasionally hit a case where it skips a second because the `setInterval` went off on either side of it. Let's do:

```suggestion
    }, 250);
```

js/app/components/timer.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <View
+      justifyContent="center"
+      flexDirection="row"
+      alignItems="center"
+      paddingHorizontal="$2"
+      paddingVertical="$1"
+    >
+      <Text fontFamily="$mono" textShadowOffset={{ width: -1, height: 1 }} textShadowRadius={3}>
+        {formatTime(elapsedTime)}
+      </Text>
```
iameli:
```suggestion
      <Text
        fontFamily="$mono"
        textShadowOffset={{ width: -1, height: 1 }}
        textShadowRadius={3}
      >
        {formatTime(elapsedTime)}
      </Text>
```

oops formatting

PR #170, opened 2025-05-13
design: redo home menu

### LINKED ISSUE #112: design: homepage & new components

espeon:
should close #112


pkg/model/segment.go
```diff
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/bluesky-social/indigo/lex/util"
```
iameli: nit: all of the bluesky code imports this as `lexutil` so i've been following that pattern
```suggestion
	lexutil "github.com/bluesky-social/indigo/lex/util"
```

js/app/features/bluesky/blueskySlice.tsx
```diff
@@ -274,6 +276,57 @@ export const blueskySlice = createAppSlice({
...
+        return await bskyAgent.getProfiles({
+          actors: actors,
+        });
+      },
+      {
+        pending: (state) => {
+          // state.status = "loading";
+        },
+        fulfilled: (state, action) => {
+          clearQueryParams();
```
iameli: These are specifically in `getProfile` for cleaning the OAuth query parameters after a successful login, shouldn't need them here

iameli (changes requested):
This looks really great! Love the pulsing live circle and the varied display of streams with the double-wide and such. Bunch of nitpicks but that's how this works:

* I think we want "real" thumbnails of what's live in the stream for the front page (as opposed to the golive post, where it can be customized) so people can see what's actually happening instead of me being stuck as this forever
image
* Speaking of that, the black background of the screenshot kind of melds into the void, would be good to have some kind of backgorund

So let's get /api/live-users migrated over to an XRPC endpoint, perhaps `place.stream.live.getLiveUsers`? You can merge in https://github.com/streamplace/streamplace/pull/181 and then follow the same patterns there to get it moved over. That endpoint is no longer needed on the live dashboard to let users know they're live so it can be changed to just return `place.stream.defs#livestreamView` records for users that are currently live; should be a lot easier to hydrate.

pkg/model/segment.go
espeon: rewrote the function elsewhere and it doesn't use lexutil 🫡

js/app/features/bluesky/blueskySlice.tsx
espeon: taken out, thank you!

iameli (approved):
woooooooo!

PR #181, opened 2025-05-17
add place.stream.live.getSegments

iameli:


espeon (approved):
hmm seems workable, will merge to my home menu branch

PR #184, opened 2025-05-17
add and fix errors reported by golangci-lint

bluwumeaway:
there are some errors being reported that require a bit more thought into how to handle. in pkg/gen/gen.go that doesn't seem like it matters too much. and aqio it should be fine to ignore the error but I could see how it can fail.
```
PKG_CONFIG_PATH=/home/blu/git/streamplace/build-linux-amd64/meson-uninstalled \
golangci-lint run -c ./.golangci.yaml
pkg/crypto/signers/signers.go:44:20: SA1019: elliptic.Marshal has been deprecated since Go 1.21: for ECDH, use the crypto/ecdh package. This function returns an encoding equivalent to that of PublicKey.Bytes in crypto/ecdh. (staticcheck)
        publicKeyBytes := elliptic.Marshal(crypto.S256(), pub.X, pub.Y)
                          ^
pkg/gen/gen.go:15:13: SA4010: this result of append is never used, except maybe in other appends (staticcheck)
                typVals = append(typVals, reflect.New(typ).Elem().Interface())
                          ^
pkg/aqio/aqio.go:68:10: Error return value of `rws.Seek` is not checked (errcheck)
        rws.Seek(0, io.SeekStart)
                ^
pkg/aqio/aqio.go:75:10: Error return value of `rws.Seek` is not checked (errcheck)
        rws.Seek(0, io.SeekStart)
                ^
pkg/atproto/util.go:18:20: SA1019: elliptic.Marshal has been deprecated since Go 1.21: for ECDH, use the crypto/ecdh package. This function returns an encoding equivalent to that of PublicKey.Bytes in crypto/ecdh. (staticcheck)
        publicKeyBytes := elliptic.Marshal(ethcrypto.S256(), ecdsaPub.X, ecdsaPub.Y)
                          ^
make: *** [Makefile:74: golint] Error 1
```


iameli:
Nice!!! I'll take a look at the weird cases, but the big one for me would be to get these checks into CI - it would be as easy as adding `$(MAKE) golint` to the `make check` options and we'll get there.

But I bet we can do better and get a GitHub action that actually gives us a visible report on what lines need to be fixed...

pkg/media/audio_smear.go
```diff
@@ -265,13 +260,11 @@ func JoinAudioVideo(ctx context.Context, seg *SegmentData, output io.Writer) err
 	errCh := make(chan error)
 	go func() {
 		err := HandleBusMessages(ctx, pipeline)
-		cancel()
```
iameli: Most of these we do need - the bus handler setting down is actually the primary sign from the gstreamer pipeline that we're done and should start shutting down

pkg/media/gstreamer.go
```diff
@@ -258,18 +262,24 @@ func (mm *MediaManager) TestSource(ctx context.Context, ms MediaSigner) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 	go func() {
+		if err := pipeline.BlockSetState(gst.StateNull); err != nil {
+			log.Log(ctx, "failed to set pipline state", "error", err)
+		}
 		<-ctx.Done()
```
iameli: Oop, definitely don't want this, we're gonna shut down the pipeline before anything happens

pkg/media/media.go
```diff
@@ -198,6 +198,7 @@ var ErrInvalidMetadata = errors.New("invalid segment metadata")
 
 func ParseSegmentAssertions(ctx context.Context, mani *manifeststore.Manifest) (*SegmentMetadata, error) {
 	ctx, span := otel.Tracer("signer").Start(ctx, "ParseSegmentAssertions")
+	defer ctx.Done()
```
iameli: This one confuses me, why do we need to wait on context.Done() after this function ends?

pkg/media/media_data_parser.go
```diff
@@ -113,15 +113,22 @@ func ParseSegmentMediaData(ctx context.Context, mp4bs []byte) (*model.SegmentMed
...
-	demux.Connect("pad-added", onPadAdded)
+	_, err = demux.Connect("pad-added", onPadAdded)
+	if err != nil {
+		return nil, fmt.Errorf("error connecting pad-add: %w", err)
+	}
 
 	go func() {
-		HandleBusMessages(ctx, pipeline)
+		if err := HandleBusMessages(ctx, pipeline); err != nil {
+			log.Log(ctx, "error handling bus messages", "error", err)
```
iameli: Nit: this is slightly incorrect - an error returning from `HandleBusMessages` usually indicates the pipeline itself exiting with an error rather than the bus handler. These could be as simple as 
```suggestion
			log.Error(ctx, "pipeline error", "error", err)
```

pkg/media/media_data_parser.go
```diff
@@ -113,15 +113,22 @@ func ParseSegmentMediaData(ctx context.Context, mp4bs []byte) (*model.SegmentMed
...
+		}
 		cancel()
 	}()
 
 	// Start the pipeline
-	pipeline.SetState(gst.StatePlaying)
+	if err := pipeline.BlockSetState(gst.StatePlaying); err != nil {
```
iameli: I'm not 100% sure that our usages of `SetState` vs `BlockSetState` are correct but let's not muck with them without a reason
```suggestion
	if err := pipeline.SetState(gst.StatePlaying); err != nil {
```

pkg/media/thumbnail.go
```diff
@@ -57,11 +57,15 @@ func Thumbnail(ctx context.Context, r io.Reader, w io.Writer) error {
 		NewSampleFunc: WriterNewSample(ctx, w),
 	})
 
-	pipeline.SetState(gst.StatePlaying)
+	if err := pipeline.BlockSetState(gst.StatePlaying); err != nil {
```
iameli, line 60:
```suggestion
	if err := pipeline.SetState(gst.StatePlaying); err != nil {
```

pkg/media/media_data_parser.go
bluwumeaway: typed the wrong thing here my bad :9

bluwumeaway:
It is using v2/most of the complaints have been fixed. Would you rather the check be a separate job/action/script that is ran or just a part of the 8 builds?

iameli:
> Would you rather the check be a separate job/action/script that is ran or just a part of the 8 builds?

A separate job would be great so it runs as fast as possible!

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4.1.7
+      - uses: actions/setup-go@v5
+        with:
+          go-version: stable
+      - name: make node
```
iameli: I was hoping we could do this without actually building the node? Like just clone the repo and immediately lint. Potentially even a GitHub action that does this?

bluwumeaway: yeah i was hoping that too but im having issues getting golangci tl run without erroring on the cgo imports

bluwumeaway: was going to look at it again later today

iameli: Hmmm perhaps `CGO_ENABLED=0` might help but yeah that makes sense

pkg/api/playback.go
```diff
@@ -86,7 +86,9 @@ func (a *StreamplaceAPI) HandleMP4Playback(ctx context.Context) httprouter.Handl
 			_, err := io.Copy(w, pr)
 			return err
 		})
-		g.Wait()
+		if err := g.Wait(); err != nil {
+			errors.WriteHTTPBadRequest(w, "request failed", err)
```
iameli, line 90:
```suggestion
			log.Error(ctx, "error writing response", "error", err)
```

pkg/api/playback.go
```diff
@@ -138,7 +140,9 @@ func (a *StreamplaceAPI) HandleMKVPlayback(ctx context.Context) httprouter.Handl
...
```
iameli, line 144:
```suggestion
			log.Error(ctx, "error writing response", "error", err)
```

pkg/api/playback.go
```diff
@@ -168,12 +172,14 @@ func (a *StreamplaceAPI) HandleWebRTCPlayback(ctx context.Context) httprouter.Ha
 		}
 		w.WriteHeader(201)
 		w.Header().Add("Location", r.URL.Path)
-		w.Write([]byte(answer.SDP))
+		if _, err := w.Write([]byte(answer.SDP)); err != nil {
+			log.Error(ctx, "error writing response", "error", err)
```
iameli, line 176:
```suggestion
			log.Error(ctx, "error writing response", "error", err)
```

PR #205, opened 2025-05-23
docs: openapi gen via lexmd

espeon:
will need to do some sort of pruning of lexicons but it runs + compiles + serves fine

right now there's an indirectly-used folder that hopefully will be deleted once I get pruning working 👍🏼


espeon:
pruning is working, should be fine to merge ~~if~~ when it builds fine

PR #220, opened 2025-05-26
fix: increase volume control tap target size/fade delay

espeon:
thinking this may be the root bug for the informal? bug report in discord that said something along the lines of "the volume control fades out when you mouse over it"


iameli (approved):
cool!

espeon:
ok trying to rebase to next one last time

PR #230, opened 2025-05-29
chat: rework rich text renderer's faceting

### LINKED ISSUE #159: chat: link facets are offset from text a little bit
This is probably going to be some UTF-16 vs UTF-8 nonsense?

Image

espeon:
uses mary's lib but I copied it over to our repo b/c we're using a different bsky types library than her package does

should close https://github.com/streamplace/streamplace/issues/159


iameli (approved):
LGTM, `git rebase origin/next` should get CI passing

PR #246, opened 2025-06-02
discord: add basic webhook integration

iameli:
[Instead of waiting on figuring out how to handle private data](https://bsky.app/profile/iame.li/post/3lpuvxbdjd22k), I'm just handling Discord integration as a "you have to run your own node" kind of deal. This will hopefully unblock the couple of users that want to do this for now.

@espeon I can't figure out how to get this below the other two pages? I have it set with `order: 30` but it doesn't seem to care.

image


js/docs/src/content/docs/guides/start-streaming/discord-hooks.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Discord Webhooks
+description: Configure Discord webhooks for livestream announcements and chat
+order: 30
```
espeon: it's definitely weird, but [according to docs](https://starlight.astro.build/reference/frontmatter/#order) this is how it's supposed to be done
```suggestion
sidebar:
  order: 30
```

espeon (approved):
lgtm! would consider changing all the other .md files to match the one I suggested for readability but it’s not 100% necessary (results are the same either way)

PR #247, opened 2025-06-02
pip: add backgrounding/enable pip on supported devices

espeon:
currently, native pip is ios-only b/c of upstream unfortunately, but i may make a patch if it isn't much of an issue


iameli:
integrated in here https://github.com/streamplace/streamplace/pull/292

PR #250, opened 2025-06-02
thumbs: change to jpeg thumbnails

### LINKED ISSUE #241: thumbnail: move to jpegs

espeon:
should close #241

one thing, haven't necessarily figured out how running go tests here works, everything I've tried results in an ambiguous result?


iameli (changes requested):
nice! let's add the test on the jpeg side per conversation

iameli (approved):

PR #251, opened 2025-06-02
discord: don't ping if you're hydrating an empty database

iameli:
On initial sync we don't wanna spam the Discord webhook


espeon (approved):
looks good to me

PR #254, opened 2025-06-04
fix: hide sidebar when embed is loaded

espeon:


iameli:
rebasing for ci improvements

iameli (approved):

PR #257, opened 2025-06-04
components: add player store

espeon:


js/components/src/player-store/player-store.tsx
```diff
@@ -0,0 +1,185 @@
...
+      console.log(`Player Event: ${eventType} at ${time}`, meta);
+    },
+
+    setUserInteraction: () =>
+      set((p) => {
+        // controls timeout
+        if (p.controlsTimeout) {
+          clearTimeout(p.controlsTimeout);
+        }
+        let controlsTimeout = setTimeout(() => p.setShowControls(false), 1000);
```
iameli, line 141: Will there be a problem/leak if this unmounts with the timeout still hanging?

js/app/components/player/controls.tsx
```diff
@@ -54,227 +53,201 @@ import {
...
-  const [sliderValue, setSliderValue] = useState(volume);
-  const { isWeb } = usePlatform();
-  const sliderRef = useRef(null);
-  const iconRef = useRef(null);
-  const lastVolumeRef = useRef(volume || 1);
-  const fadeAnim = useRef(new Animated.Value(0)).current;
-  const hideTimer = useRef<NodeJS.Timeout | null>(null);
+  const [volumeVisible, setVolumeVisible] = useState(false);
+  const [volumeSliderWidth, setVolumeSliderWidth] = useState(0);
+  const [localVolume, setLocalVolume] = useState(props.volume);
```
iameli: Stuff like this could all be zustand selectors perhaps?

js/app/components/player/provider.tsx
```diff
@@ -1,62 +1,20 @@
...
-      <PlayerContextInitializer {...props}>
+    <NewPlayerProvider>
+      <LivestreamProvider src={props.src ?? ""}>
         {props.children}
-      </PlayerContextInitializer>
-    </LivestreamProvider>
+      </LivestreamProvider>
+    </NewPlayerProvider>
   );
 }
```
iameli: Potentially overkill to have a separate file for this now, could just wrap the `player.tsx` export in a `<LivestreamProvider />` and be happy

js/app/components/player/shared.tsx
```diff
@@ -1,12 +1,10 @@
...
-  PROTOCOL_HLS,
-  PROTOCOL_PROGRESSIVE_MP4,
-  PROTOCOL_PROGRESSIVE_WEBM,
-  PROTOCOL_WEBRTC,
-} from "./props";
+
+const PROTOCOL_HLS = "hls";
+const PROTOCOL_PROGRESSIVE_MP4 = "progressive-mp4";
+const PROTOCOL_PROGRESSIVE_WEBM = "progressive-webm";
+const PROTOCOL_WEBRTC = "webrtc";
```
iameli: Import from library, yeah?

js/app/components/player/video.tsx
```diff
@@ -81,13 +96,13 @@ const VideoElement = forwardRef(
...
             if (localVideoRef.current) {
-              props.setMuted(true);
+              setMuted(true);
               localVideoRef.current.muted = true;
               localVideoRef.current
                 .play()
                 .then(() => {
-                  console.log("muted video");
-                  props.setMuteWasForced(true);
+                  console.log("muted video, forced");
```
iameli:
```suggestion
                  console.log("browser forced video to start muted");
```

js/app/components/player/video.tsx
```diff
@@ -230,44 +248,91 @@ export function HLSPlayer(props: VideoProps) {
...
+  const playerEvent = usePlayerStore((x) => x.playerEvent);
+
+  // Check WebRTC compatibility on component mount
+  useEffect(() => {
+    try {
+      checkWebRTCSupport();
+      console.log("WebRTC Player - Browser compatibility check passed");
+      logWebRTCDiagnostics();
+    } catch (error) {
+      console.error("WebRTC Player - Compatibility error:", error.message);
```
iameli, line 266: If this works we could potentially set their protocol to HLS at this point?

js/app/components/player/video.tsx
```diff
@@ -296,14 +361,14 @@ export function WebRTCPlayer(props: VideoProps) {
...
+    // if (!props.avSyncTest) {
+    //   return;
+    // }
     if (!mediaStream) {
       return;
     }
-    quietReceiver(mediaStream, props.playerEvent);
-  }, [props.avSyncTest, mediaStream]);
+    quietReceiver(mediaStream, playerEvent);
+  }, [mediaStream]);
```
iameli: I think this test is not working right now but prolly comment entirely instead of unconditionally calling the quiet receiver

js/app/components/player/video.tsx
```diff
@@ -312,13 +377,90 @@ export function WebRTCPlayer(props: VideoProps) {
...
+          </View>
+          <Text fontSize="$4" color="white" lineHeight="$1">
+            Establishing WebRTC connection ({connectionStatus})
+          </Text>
+        </View>
+      </View>
+    );
+  }
+
+  // Show error message if WebRTC is not supported
```
iameli: How did you simulate this? I want to be able to do that

iameli (approved):
Suggested some minor stuff but really really good overall!

js/app/components/player/player.tsx
```diff
@@ -1,63 +1,73 @@
...
+    setFullscreen?: (fullscreen: boolean) => void;
+  },
+) {
   return (
     <PlayerProvider {...props}>
+      <PropUpFullscreen setFullscreen={props.setFullscreen} />
       <PlayerInner {...props} />
     </PlayerProvider>
   );
 }
```
iameli: Doesn't need to be done now but when we make this an actual importable component we're gonna need to probably have a lot of `useEffect` for keeping the Zustand state in line with the user-provided props

iameli:
Oh also self-test failing, instructions for running that I should document somewhere:

```
# Tab 1:
make dev && ./build-darwin-arm64/streamplace --test-stream

# Tab 2 is whatever version of the app server you're using
pnpm run app start

# Tab 3 runs the Desktop self-test
pnpm run desktop start -- --self-test
```

Seems like something there isn't working quite right

iameli:
Having trouble with the mute button - I think maybe the slider works but the button isn't stopping sound correctly

js/app/components/player/controls.tsx
espeon: true, but we may need to have localvolume still b/c the mute should reset the global volume to 0

js/app/components/player/video.tsx
espeon: There are extensions that disable/enable it, I use firefox and this is the one i used: https://addons.mozilla.org/en-US/firefox/addon/webrtc-control/

js/components/src/player-store/player-store.tsx
espeon: possibly? i'll create an unmount function for it as I don't think unmounting is doable directly in the zustand store

espeon:
self test should be fixed and (iirc) i fixed the mute button issue earlier when I was changing the volume over to directly use zustand selectors

iameli (approved):

PR #260, opened 2025-06-06
login: force lower case, url keyboard

iameli:
"Scumb.ag" actually causes an error per https://github.com/bluesky-social/atproto/issues/3646 so may as well force all input to lower case


espeon (approved):
looks good!

PR #267, opened 2025-06-08
home: dump redux, fix long first load time

iameli:
@espeon Our first experiment into `{data, loading, error, refresh}` in Zustand — did this to speed up the first load which I think wasn't happening for a couple seconds. Not certain whether this is clever or not or if we're going to need something else entirely. (Is this... what Tanstack Query is for?)

```typescript
    setLiveUsers: (opts: {
      liveUsers?: PlaceStreamLivestream.LivestreamView[];
      liveUsersLoading?: boolean;
      liveUsersError?: string | null;
      liveUsersRefresh?: number;
    }) => {
      set({
        ...opts,
      });
    },
```

### DIFF
--- a/js/app/features/streamplace/streamplaceSlice.tsx
+++ b/js/app/features/streamplace/streamplaceSlice.tsx
@@ -202,61 +202,6 @@ export const streamplaceSlice = createAppSlice({
       },
     ),
 
-    pollSegments: create.asyncThunk(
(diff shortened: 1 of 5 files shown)


espeon:
i think that this is one of the things tanstack query or some other SWR tool would definitely be good for tbh

PR #271, opened 2025-06-09
build: try earlier mac sdk version

iameli:


iameli:
this worked but didn't work consistently, https://github.com/streamplace/streamplace/pull/273 did

PR #272, opened 2025-06-09
build: try using apple clang

iameli:


iameli:
https://github.com/streamplace/streamplace/pull/273 is working, using that

PR #280, opened 2025-06-11
player: mobile stream component

espeon:
This is a 'mobile stream' tiktok/ig live-esque component


js/components/src/components/mobile-player/video.native.tsx
```diff
@@ -0,0 +1,356 @@
...
+    );
+  }
+
+  return (
+    <RTCViewIngest
+      mirror={ingestCamera !== "environment"}
+      objectFit={"cover"}
+      streamURL={localMediaStream.toURL()}
+      zOrder={0}
+      // width is set to 5000 to ensure it fills the screen
```
iameli: lies

js/components/src/components/mobile-player/webrtc-diagnostics.tsx
```diff
@@ -0,0 +1,145 @@
...
+    const getDisplayMedia = !!navigator.mediaDevices?.getDisplayMedia;
+
+    if (!getDisplayMedia) {
+      warnings.push(
+        "getDisplayMedia is not supported - screen sharing unavailable",
+      );
+    }
+
+    // Check if running over HTTPS (required for some WebRTC features)
+    if (location.protocol !== "https:" && location.hostname !== "localhost") {
```
iameli, line 92: Prolly should include `127.0.0.1` in here - our app at least doesn't ever allow `localhost` to happen https://github.com/streamplace/streamplace/blob/915d8a491cfdb3ce1f8f2d35d5c01f875ef2fe84/js/app/components/provider/provider.tsx#L16-L23

js/components/src/components/ui/button.tsx
```diff
@@ -0,0 +1,309 @@
+import { cva, type VariantProps } from "class-variance-authority";
```
iameli, line 1: oooooh I like this a lot

js/components/src/components/ui/dropdown.tsx
```diff
@@ -0,0 +1,424 @@
+import BottomSheet, { BottomSheetView } from "@gorhom/bottom-sheet";
+import * as DropdownMenuPrimitive from "@rn-primitives/dropdown-menu";
```
iameli, line 2: I assume stuff like this adds native dependencies, which is totally fine - requiring our users to add a native dependency when integrating Streamplace will not be unexpected. But we should keep in mind that we ideally wouldn't want to do that very often for upgrades; come to think of it we should probably be doing a major version revision of the `@streamplace/components` library when we do so to avoid breaking our users via OTA updates.

So in that context... do we want to maybe add the entire `@rn-primitives` library in case we want more in the future? I guess we can see how big that is. Anything else we'd want to make sure we get into the 0.7.0 release?

iameli: (That said I'm loving these component libraries, I think this is exactly how we'd want to handle this)

js/components/src/components/ui/index.ts
```diff
@@ -0,0 +1,31 @@
+// Export primitive components
```
iameli, line 1: Barrel file :upside_down_face:

js/components/src/livestream-store/stream-key.tsx
```diff
@@ -0,0 +1,95 @@
...
+        }
+      } else if (platform === "android") {
+        platform = "Android";
+      } else if (platform === "ios") {
+        platform = "iOS";
+      } else if (platform === "macos") {
+        platform = "macOS";
+      } else if (platform === "windows") {
+        platform = "Windows";
+      }
```
iameli: Would be nice to get this working on Streamplace Desktop; that'll probably come back as `chrome` or `electron` or something with this technique. IDK off the top of my head how to detect host platform on Electron, prolly with a node.js API of some sort. Worst-case we can do a build-time environment variable of some sort

"But Eli you don't really have a plan for Streamplace Desktop right now"

Okay true but it works and I wanna keep supporting it while we flesh out the product underneath it

iameli:
This is looking sooooo gooooood I really like the component library stuff. Random Android testing:

* The countdown was really slow (in a way that felt consistent with our general Android slowness). The server actually gave up on the livestream so now I'm kinda trapped in limbo
* I can't seem to "Close" this (actually it went away since I posted this, prolly also a slow Android thing)

image

js/components/src/components/ui/dropdown.tsx
espeon: Both rn-primitives and bottom sheet are all JS for some reason. Earlier I was thinking of adding zeego which actually is native. This is probably better though as it's more updateable due to JS

iameli: Oh that rules.

espeon:
Will double check on android! i haven't fully tested there just yet, that's next on my todo list actually

js/components/src/components/ui/index.ts
espeon: can't find a way around it so far 🥲

espeon:
Closing this in favour of the 0.7 branch #302, all suggestions should be implemented so far as I know

PR #281, opened 2025-06-11
hotfix: remove `cursor: pointer` + invalid dimension on iOS

espeon:
Two fixes:

- Warning on "-$1" not being a valid dimension on iOS

- Fix occasional infinite loading on iOS. So far as I can tell, we can add this back in on the switch to New Architecture eventually.

### DIFF
--- a/js/app/components/livestream/livestream.tsx
+++ b/js/app/components/livestream/livestream.tsx
@@ -267,7 +267,8 @@ export function LivestreamInner(props: Partial<PlayerProps>) {
                                 color: "$blue11",
                               }}
                               aria-label={`View @${streamerHandle} on Bluesky`}
-                              style={{ cursor: "pointer" }}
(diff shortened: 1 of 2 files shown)


js/app/components/livestream/livestream.tsx
```diff
@@ -267,7 +267,8 @@ export function LivestreamInner(props: Partial<PlayerProps>) {
                                 color: "$blue11",
                               }}
                               aria-label={`View @${streamerHandle} on Bluesky`}
-                              style={{ cursor: "pointer" }}
+                              // TODO: re-add on New Architecture
+                              //style={{ cursor: "pointer" }}
```
iameli:
```suggestion
                              // TODO: re-add on New Architecture
                              style={isWeb ? { cursor: "pointer" } : {}}
```

iameli (approved):

PR #284, opened 2025-06-13
app: enable android profiling, add chat optimizations

iameli:


js/app/components/chat/chat.tsx
```diff
@@ -326,106 +324,106 @@ function ChatMessageRow({
...
+        paddingHorizontal={isWeb ? 6 : 4} // Adjust padding for web
+        position="relative"
+        hoverStyle={{ backgroundColor: "rgba(255,255,255,0.1)" }}
+        backgroundColor={
+          currentReplyTo?.cid === message.cid
+            ? "rgba(180,180,255,0.1)"
+            : "transparent"
+        }
+        borderRadius={isWeb ? 4 : 4}
+        onPress={() => {
```
espeon, line 346: Re-adding the long press to reply-ability to mobile would be good since we're taking away slide to reply
```suggestion
        onLongPress={() => {
            if (!isWeb) {
               handleReply();
            }
          }
        onPress={() => {
```

espeon (approved):
On mobile i would add back the long press to reply, but it looks good to me past that!

This issue on RNGH seems applicable to this if one wanted to track it: https://github.com/software-mansion/react-native-gesture-handler/issues/3344

PR #289, opened 2025-06-15
fix: throw if model repo is nil in discord integration

espeon:
I feel like the 'actual' 'issue' is that we don't fetch repos if we don't have them locally (unsure if this is intended or not), but this should do in the time being


iameli:
@espeon your check is good but i added https://github.com/streamplace/streamplace/pull/289/commits/931f098c5278fe9fb4f7d12c17baaad892c48b25 on top - turns out we did have the repo but I just referenced the wrong non-synced one :facepalm:

espeon:
Ahh, I see now, totally missed that at first. Thank you!

PR #301, opened 2025-06-21
fix: Resolve DIDs correctly

espeon:
Use the built-in function for detecting and resolving facets such that the `did` field is an actual did and not a handle.


js/components/src/livestream-store/chat.tsx
```diff
@@ -52,12 +48,15 @@ export const useCreateChatMessage = () => {
     }
 
     const rt = new RichText({ text: msg.text });
-    rt.detectFacetsWithoutResolution();
+    rt.detectFacets(pdsAgent);
```
iameli: I thought this version was async?

espeon: hmm, thought I added async to that. will update right now

js/components/src/livestream-store/chat.tsx
```diff
@@ -52,12 +48,15 @@ export const useCreateChatMessage = () => {
     }
 
     const rt = new RichText({ text: msg.text });
-    rt.detectFacetsWithoutResolution();
+    await rt.detectFacets(pdsAgent);
+
+    console.log(rt.facets);
```
iameli:
```suggestion
```

iameli (approved):

js/components/src/livestream-store/chat.tsx
```diff
@@ -42,101 +38,66 @@ export const useCreateChatMessage = () => {
...
-                  uri: msg.reply.uri,
-                },
-                parent: {
-                  cid: msg.reply.cid,
-                  uri: msg.reply.uri,
-                },
+    const rt = new RichText({ text: msg.text });
+    await rt.detectFacets(pdsAgent);
+
+    console.log(rt.facets);
```
espeon:
```suggestion
```

PR #302, opened 2025-06-21
ui: mobile player (0.7)

### LINKED ISSUE #124: app: implement from-phone streaming
[So I have a proof-of-concept here](https://github.com/streamplace/streamplace/pull/262) but it needs a lot of things. This will be a good opportunity to implement an actual component in the `@streamplace/components` library that will get used both in the Streamplace app and externally. We don't need to throw away the existing player quite yet but probably we'll continue to evolve this until that happens.

Features that this needs:
* Camera selector. It's possible that dynamically switching front/back facing cameras is blocked by dynamically handling tracks on the backend, and that's fine — for now we can just reestablish the PeerConnection entirely when the camera changes and iterate off of that.
* Chat overlayed on top of the stream like TikTok/Instagram live.
* Better indication of whether the stream is live/"healthy" or not — the best way to be certain is whether you've gotten a recent segment from the websocket
* New, mobile-first interface for creating a livestream record.

Requirements for the component library:
* No Tamagui; I just don't think it's realistic for us to require our users to bring a Tamagui theme. I wanna dump it.
* That means we'll need to implement a couple reusable menu components here with an eye toward replacing the "gear menu" that we have in the current player.

Things out of scope for the moment:
* Landscape streaming — we'll get there eventually but my inclination is to focus on polishing portrait for the first version.

espeon:


iameli:
@espeon CI error is a little opaque but I think this means these packages are used in the library without being declared as package.json dependencies 
image

iameli:
One thought - the preview screen probably shouldn't crop out anything you're streaming in the background, even if that means we have to do some kind of `contain` instead of `cover` thing and blank out the bottom of the window

image

iameli:
image

Looks like the keyboard isn't avoiding chat on iOS (working on Android tho) (edit: confirmed fixed)

espeon:
The keyboard works fine on my phone, but it could possibly be some weird layout issue. I've since set that to be a fixed height and not a ratio, and I'm thinking that should help.

espeon:
Will rebase, then it should be ready for merging
edit: forgot the chat box but that's simple and i already have it mostly done

js/docs/src/content/docs/components/custom_ui.md
```diff
@@ -0,0 +1,245 @@
+---
+title: Creating your own player UI
+description:
+  How to set up your player UI with components from @streamplace/components.
+---
+
+# Building a Custom Player UI
```
iameli, line 7: title is already there 
```suggestion
```

js/docs/src/content/docs/components/custom_ui.md
```diff
@@ -0,0 +1,245 @@
...
+  );
+}
+```
+
+## Creating Your Own UI
+
+### 1. Set Up the File
+
+Create a new file for your custom UI component, for example, `MyCustomUi.tsx`,
+inside `streamplace/js/app/components/mobile/`.
```
iameli, line 154: Probably not, right?

iameli (approved):
Looking really great! Tracking some stuff to fix but let's get this merged in and do smaller PRs on top of it, this is getting big

PR #311, opened 2025-06-23
patch: webrtc fork

espeon:
Looks like the webrtc fork is calling 'husky', and in CI, it seems to act odd when you get right from codeload.github.com


iameli:
huh okay, neat

iameli:
Part of the NPM spec I guess: https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts

> NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.

I guess I get what they're going for there, still a pain tho

iameli (approved):

PR #327, opened 2025-06-27
app: add deep linking capabilities

iameli:
@espeon Any other exclusions you can think of?

```
[]string{"*", "NOT /docs/*", "NOT /api/*", "NOT /xrpc/*", "NOT /oauth/*"},
```


espeon:
there’s well-known but afaik that’s it

PR #328, opened 2025-06-27
home: change card image from 'stretch' to 'contain'

### LINKED ISSUE #314: thumbnails: what if less wide though
Look at how wide I am!

Image

Wow that's wide. We should prolly do something else for portrait streams though.

espeon:
should fix #314 😭


iameli:
rip wide eli

PR #331, opened 2025-06-27
chat: rework layout so it's much more useable on mobile

### LINKED ISSUE #329: Android: chat button is squooshed off to the side

espeon:
Fixed #329 

- Chat is now absolute when needed 
- Set max width to 100% 
- TextArea is now flexible and not set at 100%


iameli (approved):
Looking good - I had one moment where I had weirdly-sized chat box but haven't been able to reproduce

PR #335, opened 2025-06-28
app: fixes for local chat + backwards chat in web

iameli:
Few changes:

* Fix for local chat showing up (and backwards chat in the web view)
* Fix my camera-swapping thing breaking stuff (via the `ingestLive` property that should control when we do and don't go live)
* Refactored some stuff around chat and memoized the chat lines. Chat is SO FAST now :fast_forward:

### DIFF
--- a/js/components/src/components/chat/chat.tsx
+++ b/js/components/src/components/chat/chat.tsx
@@ -77,7 +77,6 @@ const ChatLine = memo(({ item }: { item: ChatMessageViewHydrated }) => {
             setReply(item);
           }
           if (r === (Platform.OS === "android" ? "left" : "right")) {
-            console.log("setting mod view");
             setModMsg(item);
           }
           // close this swipeable
--- a/js/components/src/components/chat/chat.tsx
+++ b/js/components/src/components/chat/chat.tsx
@@ -1,11 +1,6 @@
 import { Reply, ShieldEllipsis } from "lucide-react-native";
-import { ComponentProps, useCallback, useMemo, useRef, useState } from "react";
-import {
-  FlatList,
-  ListRenderItemInfo,
-  Platform,
-  Pressable,
-} from "react-native";
+import { ComponentProps, memo, useRef } from "react";
+import { FlatList, Platform, Pressable } from "react-native";
 import Swipeable, {
   SwipeableMethods,
 } from "react-native-gesture-handler/ReanimatedSwipeable";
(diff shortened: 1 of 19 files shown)


js/components/src/components/chat/chat.tsx
```diff
@@ -50,6 +51,48 @@ function LeftAction(prog: SharedValue<number>, drag: SharedValue<number>) {
...
+        overshootFriction={9}
+        ref={(ref) => {
+          swipeableRef.current = ref;
+        }}
+        onSwipeableOpen={(r) => {
+          if (r === (Platform.OS === "android" ? "right" : "left")) {
+            setReply(item);
+          }
+          if (r === (Platform.OS === "android" ? "left" : "right")) {
+            console.log("setting mod view");
```
espeon:
```suggestion
```

js/components/src/components/chat/chat.tsx
```diff
@@ -50,6 +51,48 @@ function LeftAction(prog: SharedValue<number>, drag: SharedValue<number>) {
...
+  const swipeableRef = useRef<SwipeableMethods | null>(null);
+  return (
+    <Pressable onLongPress={() => setModMsg(item)}>
+      <Swipeable
+        containerStyle={[py[1]]}
+        friction={2}
+        enableTrackpadTwoFingerGesture
+        rightThreshold={40}
+        renderRightActions={Platform.OS === "android" ? undefined : RightAction}
+        renderLeftActions={Platform.OS === "android" ? undefined : LeftAction}
```
espeon, line 70: Unsure how mod/admin perms will be done in the future so it might be best to disable this for now? or restrict it behind if the user is the current streaming one like in the current impl

js/components/src/components/mobile-player/ui/input.tsx
```diff
@@ -1,3 +1,4 @@
+import { useKeyboardSlide } from "@streamplace/components/src/hooks";
```
espeon: this is really messy with the compilation step afaik on CI so barrel files and exporting from the base are unfortunately the best option here?

espeon (changes requested):
Good so far, but had some concerns about a few things haha. Also glad there was no difference from my impl and the naive impl lol

js/components/src/components/mobile-player/ui/input.tsx
iameli: this seems to work https://github.com/streamplace/streamplace/pull/335/commits/891768b6638ecb26008b518de0f7db859bb9c885

PR #342, opened 2025-06-30
sidebar: fix 'cyclic dependency' error on web

espeon:


iameli (approved):
lol yep that is better than my attempt where i accidentally stringified the object

PR #360, opened 2025-07-01
mobile: update styles and chat panel, + add viewer count display

### LINKED ISSUE #353: components: add live user count to mobile ui

espeon:
Should close #353 + some more nits i found

### DIFF
--- a/js/components/src/hooks/useSegmentDimensions.tsx
+++ b/js/components/src/hooks/useSegmentDimensions.tsx
@@ -1,0 +1,38 @@
+import { useEffect, useRef, useState } from "react";
+import { useLivestreamStore } from "../livestream-store";
+
+export function useSegmentDimensions() {
+  const latestSegment = useLivestreamStore((x) => x.segment);
+  const prevSegmentRef = useRef<typeof latestSegment>();
+  const prevRatio = useRef<{ height: number; width: number } | null>(null);
+
+  // Dummy state to force update every second
+  const [, setNow] = useState(Date.now());
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      setNow(Date.now());
+    }, 1000);
+    return () => clearInterval(interval);
+  }, []);
+
+  useEffect(() => {
+    // if the latestSegment is different, update the ref
+    if (latestSegment && prevSegmentRef.current !== latestSegment) {
+      prevSegmentRef.current = latestSegment;
+    }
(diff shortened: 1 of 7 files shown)


js/components/src/hooks/useSegmentDimensions.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export function useSegmentDimensions() {
+  const latestSegment = useLivestreamStore((x) => x.segment);
+  const prevSegmentRef = useRef<typeof latestSegment>();
+  const prevRatio = useRef<{ height: number; width: number } | null>(null);
+
+  // Dummy state to force update every second
+  const [, setNow] = useState(Date.now());
+
+  useEffect(() => {
```
iameli: Shouldn't be necessary IIUC? That `useLivestreamStore` is going to refresh this every time the segment changes anyway

js/components/src/hooks/useSegmentDimensions.tsx
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useRef, useState } from "react";
+import { useLivestreamStore } from "../livestream-store";
+
+export function useSegmentDimensions() {
+  const latestSegment = useLivestreamStore((x) => x.segment);
+  const prevSegmentRef = useRef<typeof latestSegment>();
+  const prevRatio = useRef<{ height: number; width: number } | null>(null);
```
iameli: What is it for?

iameli (changes requested):
Couple Qs

js/components/src/hooks/useSegmentDimensions.tsx
espeon: You're right, will remove that

js/components/src/hooks/useSegmentDimensions.tsx
espeon: The hook?  is for measuring the reported dimension of the video as the player can be inaccurate at times

The imported hooks are a remnant of the template (useSegmentTiming) i used (will remove)

iameli: I meant `prevRatio` specifically, it doesn't seem to be used anywhere

espeon: I was planning on tracking changes but then found out that was completely unnecessary

iameli (approved):
LGTM

PR #379, opened 2025-07-03
golive: disallow streaming if streaming from other device

### LINKED ISSUE #334: mobile stream: show error if you're already live elsewhere
Right now it just assumes you're live already and gets confused

espeon:


js/app/components/mobile/player.tsx
```diff
@@ -1,16 +1,101 @@
...
+    if (props.ingest && userIsLive && isStreamingElsewhere === null) {
+      setIsStreamingElsewhere(true);
+    } else if (props.ingest && userIsLive === false) {
+      setIsStreamingElsewhere(false);
+    }
+  }, [userIsLive]);
+
+  const navigation = useNavigation();
+
+  console.log(isStreamingElsewhere);
```
iameli:
```suggestion
  useEffect(() => {
    if (props.ingest && userIsLive && isStreamingElsewhere === null) {
      setIsStreamingElsewhere(true);
    } else if (props.ingest && userIsLive === false) {
      setIsStreamingElsewhere(false);
    }
  }, [userIsLive]);

  const navigation = useNavigation();

```

iameli (approved):
LGTM, just removed a couple console.log

Issue #380, opened 2025-07-03
v0.7.2 crashing when "settings" open after OTA update


iameli:
v0.7.2 seems to crash when opening the "Settings" page after OTA updating from v0.7.0.

A production build of v0.7.2 doesn't seem to reproduce the issue.

<img width="1624" alt="Image" src="https://github.com/user-attachments/assets/33405c79-26f5-4d39-b464-33304ae53bda" />

Crash log: https://storage.googleapis.com/streamplace-crap/b2157c4e-8ebd-4c94-af09-e53d9c235490/Streamplace-2025-07-02-201349.ips

[Error is the same as this StackOverflow post](https://stackoverflow.com/questions/79192259/react-native-track-player-ios-troubleshooting-building-100/79192631): `com.facebook.react.runtime.JavaScript (9): EXC_BAD_ACCESS`

iameli:
I replaced <Settings /> with an empty component and the issue still reproduced. The heck?

<img width="431" alt="Image" src="https://github.com/user-attachments/assets/ded20030-8623-4fdf-9767-1a05f8274eef" />

iameli:
The last logging from the app AFAICT:

```
(input_volume_stats_reporter.cc:98): Will not log any `WebRTC.Audio.Apm.AppliedInputVolume.*` histogram stats.
(input_volume_stats_reporter.cc:98): Will not log any `WebRTC.Audio.Apm.RecommendedInputVolume.*` histogram stats.
(objc_audio_device.mm:86): Init: Failed to initialize audio device
(objc_audio_device.mm:86): Init: Failed to initialize audio device


#
# Fatal error in: ../../media/engine/adm_helpers.cc, line 39
# last system error: 0
# Check failed: 0 == adm->Init() (0 vs. -1)
# Failed to initialize the ADM.
```

iameli:
There's really not much obvious in here that could be causing a new problem https://github.com/streamplace/streamplace/compare/v0.7.1...v0.7.2

iameli:
Now I've reproduced it on v0.7.1 on an App Store build. But then I restart the app and everything's fine. IDGI.

PR #381, opened 2025-07-03
Revert "packages: fix import urls"

iameli:
From testing this is apparently causing https://github.com/streamplace/streamplace/issues/380 but I have _zero_ idea how


iameli:
Actually nope maybe it's just random

PR #382, opened 2025-07-03
golive: increase thumb retries, add loading overlay

### LINKED ISSUE #376: mobile streaming: wait longer for thumbnail to get created
And maybe proactively show a "waiting for thumbnail creation" message

espeon:
Loading overlay is for after we count down, and before we get back from the server that it recieved our stream record

### DIFF
--- a/js/components/src/streamplace-store/stream.tsx
+++ b/js/components/src/streamplace-store/stream.tsx
@@ -18,7 +18,7 @@ const uploadThumbnail = async (
       while (
         thumbnail.data.blob.size === 0 &&
         customThumbnail.size !== 0 &&
-        tries < 3
+        tries < 12
       ) {
         console.warn(
           "Reuploading blob as blob sizes don't match! Blob size recieved is",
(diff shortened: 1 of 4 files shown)


js/components/src/streamplace-store/stream.tsx
```diff
@@ -18,7 +18,7 @@ const uploadThumbnail = async (
       while (
         thumbnail.data.blob.size === 0 &&
         customThumbnail.size !== 0 &&
-        tries < 3
+        tries < 12
```
iameli: With this I'm concerned about what happens when they navigate away from the stream before the upload - are we gonna display an error message 15 seconds later? Would be nice to have some kind of cancellation here (which you could probably do by making this into a `useUploadThumbnail` hook that knows when it's unmounted, ig?)

iameli (changes requested):
Looks reasonable but haven't tested. We should prolly figure out a way to cancel thumbnail upload

js/components/src/streamplace-store/stream.tsx
espeon: hmm, good point here. this is definitely doable

iameli (approved):
LGTM

PR #385, opened 2025-07-06
chat: on mobile, use a larger touch area instead of hitslop

### LINKED ISSUE #361: mobile: chat dragging on android is kinda hard

espeon:
Looks like something's overriding the hitslop on web/android and not on iOS, so switching to simply having a bigger touch area. Shown in the image below
image

fixes https://github.com/streamplace/streamplace/issues/361


iameli (approved):
yep LGTM

PR #386, opened 2025-07-08
player: start work on responsive design New Player

### LINKED ISSUE #352: player: finish consolidating old and new players

espeon:


espeon:
Still a few things left to do (but not much!) before we reach parity here

iameli (commented):
Something was funny on my initial load; stream wouldn't start. Worked aver resizing the window a little bit

iameli (changes requested):
This is looking really really good!

* [x] On my first load the stream wasn't starting, goes away if I resize the window to restart the connection
* [x] Seems like with the above I wasn't getting any kind of loading status
* [x] I think we're missing the PiP button (from the other player too)
* [x] The `@` and smilie are a little cut off on my iPhone
image


* [x] I'm not logged in on that screenshot but there's still a chat box - need some kind of prompt to go log in there
* [x] Go ahead and promote this player to the main route - we can preserve the old one at `/legacy-player` or something for a moment (but it'll be great to get it out of there and start the process of completely removing tamagui)
* [x] My chat messages seem to be showing up twice - I checked and it does appear to be two different records getting created
image

* [x] The offline experience doesn't seem great, I'm just getting a black screen right now (and frozen stream on iPhone)

image

iameli:
* [x] I was pressing the emoji and mention button and found a `<div>` somewhere lol 

image

* We should get the self-test things moved over to the new player or they're not testing anything meaningful anymore. [You should be able to do this easily enough by just porting over the `Multi` component](https://github.com/streamplace/streamplace/blob/2fa6f862cf0f7d31ebddc9629c929d01bb4f520e/js/app/src/screens/multi.tsx) which is what backs that grid of test players

espeon:
related: I also need to switch over the embed oomponent as well

also:
- [x] debug component

iameli:
This is really close!

* [x] Think we're still missing a PiP button on desktop - Firefox has its own but Chrome-y browsers need one
* [x] This button seems to not do anything
image

* [x] Another nit: the controls in the lower-right aren't clickable when the big mute button is present:

image

espeon:
pip would require a rebase (afaik?) but will get it done sometime today 👍🏼

espeon:
last thing: loading experience is a little subpar, will hack on that for a bit

iameli:
* [x] My name shows up briefly as white when I chat:

image

iameli:
This showing up for me on iPhone at the moment but I don't trust my Devplace build so I'll try fresh

image

iameli:
Repro'd on Android too, looks real

iameli (approved):
Let's gooooooo! More that we can polish but let's get it in next

Squash merge plz

PR #387, opened 2025-07-09
devx: switch to changesets for changelog and js versioning

### LINKED ISSUE #348: devx: implement a proper changelog
prolly https://github.com/changesets/changesets

espeon:


iameli (changes requested):
We talked through this but just capturing here:

* @streamplace/app needs to rev with every release of the repo, all of the rest of the JS packages can rev only when appropriate
* Ideally we'd have the ability to capture Go-only stuff in the changelog too, not sure how  hard that would be to make work well

espeon:
might be more worth it to redo atp

PR #388, opened 2025-07-09
player: add reloading test + reloading fix?

### LINKED ISSUE #350: streams seem to be not auto-reloading
We just get an offline countdown that says zero over and over :(

We should add an automated end-to-end test on this, added to this list, where a stream runs and then stops for 10 seconds and then plays again:

https://github.com/streamplace/streamplace/blob/5ba9e883d66902f99cfb20fa4ec10f66535e9274/js/desktop/src/tests/test-runner.ts#L13-L16

espeon:
Still don't know what the root cause of this is, but our way of detecting it broke. Looks like one previously used a 'ref' as a key, [which doesn't work](https://react.dev/learn/referencing-values-with-refs).


iameli (approved):
Nice, just tested this successfully! You've got a merge conflict in there somehow but LGTM

espeon:
Will merge once this builds, then.

espeon:
🎊

PR #389, opened 2025-07-09
player: reporting interface

espeon:
Adding a simple reporting interface that doesn't go anywhere just yet.


iameli:
Combined with https://github.com/streamplace/streamplace/pull/422 for testing

PR #392, opened 2025-07-12
golive: correctly resolve facets on bluesky post

### LINKED ISSUE #391: livestream: bluesky post maker does not properly facet titles

espeon:
should close #391


iameli (approved):
Yep looks good!

PR #393, opened 2025-07-12
chat: add 'hide message' feature

### LINKED ISSUE #355: moderation: add streamer "hide message" support for chat messages via post-gating

espeon:


js/components/src/components/chat/mod-view.tsx
```diff
@@ -62,52 +77,61 @@ export const ModView = forwardRef<ModViewRef, ModViewProps>(() => {
...
-                  onPress={
-                    onDeleteMessage
-                      ? () => onDeleteMessage(modMessage)
-                      : undefined
-                  }
+            {/* TODO: Checking for non-owner moderators */}
+            {channelId === handle && (
+              <DropdownMenuGroup title={`Moderation actions`}>
+                <DropdownMenuItem
+                  disabled={message.author.did === agent?.did}
```
iameli: I mean. I'm not sure the use case for hiding your own messages but... why not, I guess?

iameli (changes requested):
Looks exactly right but I can't get the menu to pop up :(

espeon:
> Looks exactly right but I can't get the menu to pop up :(

Should be fixed, I was using a weird hook that required navigator or something odd like that.

iameli (approved):
Looking great now! One nit on the mod view generally: I'd kind of expect to be able to tap off the mod view by clicking the dead space above the thing; dragging the thing all the way down is a bit less familiar to me. NBD (And plausibly not new) though

iameli (approved):
Yep, looks great now!

PR #398, opened 2025-07-16
make internal API url configurable for live subcommand

espeon:
Important for dockerized deployments of Mistserver, without directly using host networking mode

e.g. if one is setting up mistserver elsewhere, one can set the internal API url to the specific host IP (assuming they have the firewall set up correctly), or to something like host.docker.internal.

### DIFF
--- a/pkg/cmd/streamplace.go
+++ b/pkg/cmd/streamplace.go
@@ -87,11 +87,21 @@ func start(build *config.BuildFlags, platformJobs []jobFunc) error {
 	}
 
 	if len(os.Args) > 1 && os.Args[1] == "live" {
-		if len(os.Args) != 3 {
-			fmt.Println("usage: streamplace live [stream-key]")
+		cli := config.CLI{Build: build}
+		fs := cli.NewFlagSet("streamplace live")
+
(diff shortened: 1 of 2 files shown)


pkg/cmd/streamplace.go
```diff
@@ -87,11 +87,21 @@ func start(build *config.BuildFlags, platformJobs []jobFunc) error {
 	}
 
 	if len(os.Args) > 1 && os.Args[1] == "live" {
-		if len(os.Args) != 3 {
-			fmt.Println("usage: streamplace live [stream-key]")
+		cli := config.CLI{Build: build}
+		fs := cli.NewFlagSet("streamplace live")
+
+		err := cli.Parse(fs, os.Args[2:])
```
iameli, line 93: This is fine but just a note to myself: I'm about to make `--public-host` a required parameter, will need to make sure that gets handled here

iameli (approved):
This makes sense and is an improvement but I'm generally unhappy with the state of the RTMP sidecar situation, maybe I need to go throw rocks at MediaMTX's bedroom window until they listen to me 💢

PR #407, opened 2025-07-21
docs: add self hosting directions + link to repo

### LINKED ISSUE #406: docs: info about how to self-host streamplace

espeon:
https://github.com/streamplace/selfhosting

### DIFF
--- a/js/docs/src/content/docs/guides/start-streaming/self-host.md
+++ b/js/docs/src/content/docs/guides/start-streaming/self-host.md
@@ -1,0 +1,193 @@
+---
+title: Self-Hosting Streamplace
+description:
+  Self-host a personal copy of Streamplace using Docker and systemd. For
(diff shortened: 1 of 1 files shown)


iameli:
I don't really wanna recommend doing it like this :cry: Too many moving parts, too much conceptual complexity - would prefer to do https://github.com/streamplace/streamplace/issues/412 over Caddy and https://github.com/streamplace/streamplace/issues/403 over the Mist sidecar and https://github.com/streamplace/streamplace/issues/411 over telling people to write one one. We can keep your repo in there but if we put this in the docs people will want to run it like this and we'll have to tell 'em to change everything later

Those tickets I listed can be super high priority though! Kinda the point of the next big multi-node push

espeon:
hmm fair enough, may as well look at doing those instead and revisit this idea in the future

PR #410, opened 2025-07-22
desktop: move to unforked maker-appimage

iameli:
Our old forked version seemed to be breaking CI. Also it seems like it works better now


iameli:
yesss finally

PR #414, opened 2025-07-23
add lex-reference to gitignore

espeon:


espeon:
hmm, probably won't be able to do much with this bc of the order that everything is meant to be built unfortunately.

PR #416, opened 2025-07-24
api: add certmagic

espeon:
still no clue how to test this honestly. maybe replacing my current SP server with this and not going through caddy would work


iameli (changes requested):
Fun and easy! The only thing I'm looking for, I think: I wanna store the certs in the database instead of on the filesystem so we can share them between nodes eventually. CertMagic support something like that?

espeon:
yea definitely, for now i can create a fs-based saving plugin thing and we can edit it later on to be db-based later on

espeon:
will probably close this for now?

PR #419, opened 2025-07-25
theme: force dark on local themeproviders

### LINKED ISSUE #418: chat: force dark theme for text
Image

https://bsky.app/profile/xan.lol/post/3luqp5vboxs2c 

Looks like only the text here is set to the system theme, everything else is forced dark (not everyone uses dark theme :()

espeon:


iameli (approved):
yeah!

PR #420, opened 2025-07-25
implement basic stream problem detection

iameli:
@espeon Tossing this over to you. I got it this far, but it should presumably be integrated with the new player in some capacity as well. (Whole live dashboard needs some thought, I reckon.)

image

ARGUABLY this should be on the server so clients don't need to be smart about it, but this was a smaller change.


espeon:
mhm, for now the best places to put it is on the current dashboard and probably the mobile golive page? though i don't know if it would help much on the latter

espeon (commented):
good stuff, def needs a UI pass (i will do the ui pass) and that's about it

iameli:
> and probably the mobile golive page? though i don't know if it would help much on the latter

haha yeah. "your stream sucks, use a better client scrub"

espeon:
i don't think i'm 100% done with the current ui (not 100% sure on the modal part), but i'm thinking the 'final' ui will be similarly laid out

PR #421, opened 2025-07-25
desktop: add server restart test for playback recovery

iameli:
Should catch the "we just upgraded Streamplace" case that's kinda common!

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -57,29 +57,29 @@ jobs:
           path: ./bin/
           retention-days: 1
 
-  windows-selftest:
-    name: windows-selftest
-    runs-on: windows-latest
-    needs: linux
-    steps:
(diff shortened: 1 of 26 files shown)


iameli:
wait wtf it passed, it was not supposed to pass 😠

iameli:
@espeon Does https://github.com/streamplace/streamplace/pull/421/commits/d13ea62b05ff9e0601caf117ec8041b69e14fefc make sense to you? I'm going for two states:

"Playing" exclusively means we're actively showing the user something. The best measure of this is the `getStats()` stuff in the WebRTC code, that's how we know we're actually getting frames

"Offline" is just a derived state off of segments; if you haven't seen a segment in 10 seconds we show the countdown

espeon:
Yeah, this definitely makes checking for player states so much simpler

PR #429, opened 2025-08-01
segment: add size in bytes to segment lexicon

### LINKED ISSUE #427: add "size" field to place.stream.segment
This would let us calculate bitrate for showing on a dashboard etc

espeon:
should close #427

### DIFF
--- a/pkg/model/segment.go
+++ b/pkg/model/segment.go
@@ -31,10 +31,11 @@ type SegmentMediaData struct {
 	Video    []*SegmentMediadataVideo `json:"video"`
 	Audio    []*SegmentMediadataAudio `json:"audio"`
 	Duration int64                    `json:"duration"`
+	SizeB    int                      `json:"sizeB"`
 }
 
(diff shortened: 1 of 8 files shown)


pkg/model/segment.go
```diff
@@ -30,10 +30,11 @@ type SegmentMediaData struct {
 	Video    []*SegmentMediadataVideo `json:"video"`
 	Audio    []*SegmentMediadataAudio `json:"audio"`
 	Duration int64                    `json:"duration"`
+	SizeB    int                      `json:"sizeB"`
```
iameli: Probably could live with just calling this `size` for consistency unless that's reserved somewhere or something

iameli (approved):
Suggested that `size` instead of `sizeB` is probably fine but NBD either way

pkg/model/segment.go
espeon: will do haha, i thought sizeb was slightly more descriptive but i guess it doesn't really matter as size in lexicon says bytes in docs

PR #436, opened 2025-08-04
chat: link to right bsky account in message options

espeon:


iameli (approved):
Aw sad I thought it was hardcoded to link to MY account every time :(

LGTM

PR #440, opened 2025-08-06
fix: unscope functions in dropdown

### LINKED ISSUE #434: player: rn bottom view crashes on load

espeon:
The dropdown is portaled to the top of the app on mobile (it's actually a sheet on mobile), so scoped data like react contexts won't work properly.


iameli (approved):
Okay, makes sense, let's definitely ship this and unbreak. In general, though, is there a more generic way of handling this kind of thing? Feels like "passing React contexts through portals" that would have a generic solution somewhere

espeon:
i think a context bridge would be pretty decent in this case: https://github.com/pmndrs/its-fine#usecontextbridge

Issue #195, opened 2025-05-20
app: dump bluesky appview, index bluesky profiles ourselves


iameli:
So this ticket was originally about https://github.com/streamplace/streamplace/blob/cbd5e428ef0b57cc430929a6eacbc6ead45e8f9f/js/app/features/bluesky/blueskySlice.tsx#L372-L372

And we should still fix it for that but also https://tangled.sh/@stream.place/streamplace/issues/1#comment-804897 indicates that [our use of `app.bsky.actor.getProfile` in the login flow](https://github.com/streamplace/streamplace/blob/cbd5e428ef0b57cc430929a6eacbc6ead45e8f9f/pkg/spxrpc/app_bsky_actor.go#L19-L24) to get users' avatars isn't like universally guaranteed to work, it relies on the PDS connecting to Bluesky's public AppView. So if we start indexing those profiles ourselves we can handle that case too.

espeon:
I'll also note we use the appview when generating OG images here:
https://github.com/streamplace/streamplace/blob/2c8a057afba9cb815034580b60b351b2f05308bf/pkg/api/og.go#L430-L432
Personally i'm a big proponent of bringing your bsky info in, but having our own profile info lexicon. Makes it a little easier handing people who may not use bsky for one reason or another.

iameli:
Indexing is done now - there are still a couple places where we call the Bluesky AppView that we should track down and replace

PR #448, opened 2025-08-07
feat: og image generation

### LINKED ISSUE #215: OpenGraph preview cards always show final frame of previous stream, which may not be useful
My streamplace account is https://stream.place/dryad.technology . I did a stream on Friday. This site loads the current "preview card", such as is used by Discord or Bluesky:

https://www.opengraph.xyz/url/https%3A%2F%2Fstream.place%2Fdryad.technology

Currently stream.place freezes the preview card at the moment the stream cuts off, which means my stream summary image is the final frame of the last stream. It might be interesting if you gave us the *option* of doing this, but this is a difficult thing to control and it potentially may not be a very good image. For example, my last stream ended with me doing some live testing with your chat API, and in the final frame— I think I was focusing on navigating the OBS interface very hard— I am making a very silly face.

opengraph.xyz screenshot

stream.place image

*Expected behavior*: You should use the Stream thumbnail, or should allow us to upload a "permanent thumbnail" which appears between streams. (I'm not sure but I think the stream thumbnail is cleared after each stream)

espeon:
should close #215 by adding OG images. maybe we could include a dimmed/tinted screenshot of the most recent preview image if one is streaming if we want to be more obvious about that without explicitly saying "hey xyz is streaming"? also we may want to cache these for a period of time as well.

Preview card: 
image


espeon:
oh whoops i need to replace the current images first

pkg/api/og.go
```diff
@@ -0,0 +1,462 @@
...
+		log.Debug(ctx, "OG image generated and served", "username", username, "code", code)
+	}
+}
+
+func downloadImage(url string) ([]byte, error) {
+	if url == "" {
+		return nil, errors.New("empty URL provided")
+	}
+
+	resp, err := http.Get(url)
```
iameli: Should use `aqhttp` here to get our user-agent and stuff, an example: https://github.com/streamplace/streamplace/blob/395eb2f39e8a1befe6036836497089b1a60f7d49/pkg/integrations/discord/send-chat.go#L51-L61

pkg/api/og.go
```diff
@@ -0,0 +1,462 @@
...
+	// First try to resolve via internal DB
+	repo, err := a.ATSync.Model.GetRepoByHandleOrDID(username)
+	if err != nil {
+		// log.Warn(ctx, "failed to resolve via ATSync, trying username directly", "username", username, "error", err)
+		// // Fall back to using the username directly
+		// actor = username
+		// if !strings.HasPrefix(username, "did:") && !strings.Contains(username, ".") {
+		// 	actor = username + ".bsky.social"
+		// }
+
```
iameli:
```suggestion
```

pkg/api/og.go
```diff
@@ -0,0 +1,462 @@
...
+	} else {
+		return nil, fmt.Errorf("no repo found for username: %s (%w)", username, ErrUserNotFound)
+	}
+
+	// TODO: check if actor is restricted
+
+	// Fetch full profile from Bluesky public API
+	client := &xrpc.Client{
+		Host: "https://public.api.bsky.app",
+	}
```
iameli: Fine for now but please add a reference to this in https://github.com/streamplace/streamplace/issues/195 so we know to fix this usage too

pkg/api/og.go
```diff
@@ -0,0 +1,462 @@
...
+// InvalidateOGImageCache removes a specific user's OG image from cache
+func (a *StreamplaceAPI) InvalidateOGImageCache(username string) {
+	cacheKey := fmt.Sprintf("og_image_%s", username)
+	a.OGImageCache.Delete(cacheKey)
+}
+
+// ClearOGImageCache removes all cached OG images
+func (a *StreamplaceAPI) ClearOGImageCache() {
+	a.OGImageCache.Flush()
+}
```
iameli: For why?

pkg/api/api.go
```diff
@@ -192,6 +195,7 @@ func (a *StreamplaceAPI) Handler(ctx context.Context) (http.Handler, error) {
 	addHandle(apiRouter, "GET", "/api/bluesky/resolve/:handle", a.HandleBlueskyResolve(ctx))
 	addHandle(apiRouter, "GET", "/api/view-count/:user", a.HandleViewCount(ctx))
 	addHandle(apiRouter, "GET", "/api/clip/:user/:file", a.HandleClip(ctx))
+	addHandle(apiRouter, "GET", "/api/og/:username/og.jpg", a.HandleOGImage(ctx))
```
iameli: Let's ship this as a new XRPC so it's documented and stuff. Perhaps `place.stream.live.getTitleCard` or something like that? Let me know if you get in trouble with the codegen

pkg/api/api.go
```diff
@@ -99,6 +101,7 @@ func MakeStreamplaceAPI(cli *config.CLI, mod model.Model, signer *eip712.EIP712S
 		connTracker:      NewWebsocketTracker(cli.RateLimitWebsocket),
 		limiters:         make(map[string]*rate.Limiter),
 		SignerCache:      make(map[string]media.MediaSigner),
+		OGImageCache:     cache.New(5*time.Minute, 10*time.Minute), // 5min TTL, 10min cleanup
```
iameli: Cool, we should probably use this in more places

pkg/api/og.go
```diff
@@ -0,0 +1,462 @@
...
+
+		return nil, fmt.Errorf("%w: %w", ErrUserNotFound, err)
+	} else if repo != nil {
+		// Use the DID as it's the most reliable identifier
+		actor = repo.DID
+	} else {
+		return nil, fmt.Errorf("no repo found for username: %s (%w)", username, ErrUserNotFound)
+	}
+
+	// TODO: check if actor is restricted
```
iameli: Mm. Something to be conscious of in general but in this case you'd have to manually type in someone's name... seems about as hard for someone to do as going to pdsls.dev

pkg/fonts/fonts.go
```diff
@@ -0,0 +1,21 @@
+package fonts
+
+import (
+	_ "embed"
+)
+
+//go:embed AtkinsonHyperlegibleNext-Regular.ttf
```
iameli: Overkill if they're also in js/app/assets/fonts, you can get them from the App FS.

iameli (changes requested):
yayy it's not a microservice!

biggest change I asked for is moving it into an XRPC, let me know if you run into trouble with the codegen there at all

pkg/api/og.go
espeon: thought it would be nice to manage it but on reflection it's probably not needed

pkg/fonts/fonts.go
espeon: haha good point, will do

pkg/api/api.go
espeon: Not sure if XRPC is the best option here, as we're basically just serving a blob? I guess we could have it be unspecced or something but I have zero clue where to start on this this lexicon wise

pkg/api/api.go
espeon: mhm a nice little cache can help a lot

pkg/api/og.go
espeon: fair enough, i guess it's best not to overbuild when you don't need to. if this becomes a problem i guess that's the best time to deal with it

pkg/api/api.go
iameli: There's precedent for that! I thiiiiink the codegen tools even handle it and let you just return an `io.Reader` or something from the stub functions but let me know if you have problems https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/sync/getBlob.json

iameli (changes requested):
I think I like `place.stream.live.getTitleCard` or `getProfileCard` or something more than "OGImage" which is only one place where we could use these potentially. Good to go otherwise though! Looks like you might have a go.mod merge conflict

espeon:
Good point. Will probably do `getProfileCard` in this case as there's no specific livestream title in the og image

espeon:
a second opinion on this rebase would probably be good if you have time, not used to dealing with go deps myself 🥲

PR #450, opened 2025-08-09
dash: widgets-based dashboard

### LINKED ISSUE #30: Chat window on live dashboard
Ideally the chat needs to pop out into it's own window for the streamer
this would also allow the chat window to be docked in OBS
further down the road you would want to expose a chat api so client clients like Chatterino, Twitchat.fr, streamer.bot, Aitum could connect so people can multi stream.

### LINKED ISSUE #425: golive: rework dashboard
It would be nice if the LiveDashboard page was more like an actual dashboard, including widgets and such for obs.

Definitely not highest priority but would be nice to implement!

espeon:
dashboard is now a bunch of mostly self-contained widgets that can be embedded elsewhere, such as in a OBS webview.


iameli:
Cooool! Open to moving the widgets into `components`? Seems like anyone who's building embedded streaming in their app might want these things too 🤔

iameli (changes requested):
Looks really good!

* [x] Needs the option to mute, which is what you want most of the time if you're streaming
* [x] This thing wiggles when it changes the number of digits 

Image

* [x] This is inaccurate. Off by 1000x maybe? I think my test stream is around 10 Mbps

Image

* [x] I was getting some "bad" connection readings even though I'm streaming from localhost. Let's hold off on that for now... the best way to detect that IME is through dropped frames, which we can totally add to the segment record but we don't have it quite yet

* [x] I don't think we need to spend a ton of time optimizing this for a mobile view, but it probably wouldn't be too hard to put in a few breakpoints to make stuff wrap so we don't get this:

Image

* [ ] This thing seems broken:

image


I'll do a proper code review next round

espeon:
i guess that uptime thing really ties into the "what really even qualifies as a stream" thing so may just remove it for now?

PR #460, opened 2025-08-13
chat: refocus text area on submit

espeon:


iameli (commented):
Hmmmmmmmm why is it _un_focusing though? 🤔

espeon:
...good point. I think the default behaviour is to blur but looking at this again it is kind of overkill to what i could have done haha

edit: So far as I can tell what I do is sort of necessary? If you press the "send" button, i'd think you want it to act like pressing "enter" to submit, so we'd need to refocus the box if not already done

iameli (approved):
Sure, LGTM

PR #467, opened 2025-08-15
feat: add user consent metadata system with c2pa

uditvira:
Enhances the current c2pa integration in both the UI and the node with the following:
- Users can set content metadata via a UI form. This is saved to `place.stream.default.metadata` record in the user repo. This record includes information about content warnings, content rights, and distribution policy that are expressed by the user.
- Syncs the metadata record to local database
- Introduces a ManifestBuilder that handles generating the manifest used during signing, by mapping from the `place.stream.default.metadata` lexicon to the c2pa manifest.
- Allows node operators to pass in `--content-filters /path/to/yaml` file with content filtering set. An example is provided. There are two content filters currently implemented: node operators can filter by content warning or by distribution policy.
- Viewers can see content warnings and rights indicated in the UI


iameli:
Looking awesome! I'll do some code review on the backend, will ask @espeon to look at the frontend stuff

lexicons/place/stream/default/metadata.json
```diff
@@ -0,0 +1,166 @@
+{
+  "lexicon": 1,
+  "id": "place.stream.default.metadata",
```
iameli: Perhaps `place.stream.metadata.configuration`? To reflect this is the input metadata instead of the output.

lexicons/place/stream/default/metadata.json
```diff
@@ -0,0 +1,166 @@
...
+      "properties": {
+        "allowArchive": {
+          "type": "boolean",
+          "description": "Whether nodes can archive this stream."
+        },
+        "broadcastExpiry": {
+          "type": "string",
+          "format": "datetime",
+          "description": "When rebroadcast permissions expire. If not specified, there is no expiration on rebroadcast permission."
+        }
```
iameli: I think these are one and the same - if you don't want archival, set a five-minute expiry.

Also leaning toward `deleteAfter` to make the syntax here really clear

lexicons/place/stream/default/metadata.json
```diff
@@ -0,0 +1,166 @@
...
+      "description": "The content could be perceived as offensive due to the language used."
+    },
+    "nudity": {
+      "type": "token",
+      "description": "The content could be perceived as offensive due to nudity."
+    },
+    "PII": {
+      "type": "token",
+      "description": "The content contains information that can be used to identify a particular individual, such as a name, phone number, email address, physical address, or IP address."
+    },
```
iameli: I know this is coming right from NewsCodes but I'm not sure there's utility to this one. Presumably  this kind of content shouldn't be broadcast at all, not sure what they use case here would be.

pkg/config/config.go
```diff
@@ -168,6 +182,7 @@ func (cli *CLI) NewFlagSet(name string) *flag.FlagSet {
 	fs.StringVar(&cli.AndroidCertFingerprint, "android-cert-fingerprint", "", "android cert fingerprint for deep linking")
 	cli.StringSliceFlag(fs, &cli.Labelers, "labelers", "", "did of labelers that this instance should subscribe to")
 	fs.StringVar(&cli.AtprotoDID, "atproto-did", "", "atproto did to respond to on /.well-known/atproto-did (default did:web:PUBLIC_HOST)")
+	fs.StringVar(&cli.ContentFiltersPath, "content-filters", "", "path to YAML file with content filtering rules")
```
iameli: Sooo the configuration we've encouraged users to do so far has just been with environment variables, I think we'll want to stick with that for now. Yes, stuffing all of this configuration into a JSON string is ugly, but I think the remedy for that will probably be creating a proper admin configuration interface and saving it in the new stateful database I introduced here https://github.com/streamplace/streamplace/pull/456/files#diff-f30f431d4de9ce917ded96a3f035d886824f24b188769b783ef3ead0b88a9cf1

So let's just do JSON in an env var for now

pkg/media/manifest_builder.go
```diff
@@ -0,0 +1,177 @@
...
+				"label": StreamplaceMetadata,
+				"data": obj{
+					"@context": obj{
+						"dc": "http://purl.org/dc/elements/1.1/",
+						"Iptc4xmpExt": "http://iptc.org/std/Iptc4xmpExt/2008-02-29/",
+						"photoshop": "http://ns.adobe.com/photoshop/1.0/",
+						"xmpRights": "http://ns.adobe.com/xap/1.0/rights/",
+					},
+					"dc:creator": streamerName,
+					// TODO: Add the title of the livestream. This should come from the livestream record.
```
iameli, line 60: 👍

pkg/media/manifest_builder.go
```diff
@@ -0,0 +1,177 @@
...
+				},
+			},
+		},
+	}
+
+	// Add database metadata if available
+	if mb.model != nil {
+		metadata, err := mb.model.GetDefaultMetadata(ctx, streamerName)
+		if err != nil {
+			log.Warn(ctx, "ManifestBuilder: failed to retrieve metadata, using defaults", "error", err, "did", streamerName)
```
iameli: Our convention here is that `model` methods that retrieve one record return `nil, nil` when that record doesn't exist. So if we got an error it's something worse than "metadata not found" and it'd be appropriate for us to error out:

```suggestion
			return nil, fmt.Errorf("failed to retrieve metadata: %w", err)
```

pkg/media/manifest_builder.go
```diff
@@ -0,0 +1,177 @@
...
+		// Credit Line
+		if metadata.ContentRights.CreditLine != nil {
+			mani["assertions"].([]obj)[1]["data"].(obj)["photoshop:Credit"] = *metadata.ContentRights.CreditLine
+		}
+
+		// Build the license field
+		if metadata.ContentRights.License != nil {
+			// Map internal license codes to known licenses
+			var licenseCodeMap = map[string]string{
+				"place.stream.default.metadata#cc0_1__0": "http://creativecommons.org/publicdomain/zero/1.0/",
```
iameli: Constants here seems prudent, I'm totally gonna type `place.stream.default.metadata#cc0_1_0` someday and be completely unable to spot the typo

iameli (changes requested):
Left a bunch of inline comments on the Go code - I think your implementation is mostly spot-on, shipping this will just be a matter of iterating on the lexicon/C2PA schema shape per our discussion.

Oh, at some point I'd love for all the C2PA signing and verification stuff to move out of the `media` package, which is way bloated and does way too many things. It deserves its own top-level thing, and then `media` can stay focused on the GStreamer pipelines that actually muck with the video. But that's not a dealbreaker, just something to take on if you have time.

js/app/components/content-metadata-form.tsx
```diff
@@ -0,0 +1,1120 @@
...
+    value: "place.stream.default.metadata#PII",
+    label: "Personally Identifiable Information",
+  },
+  { value: "place.stream.default.metadata#sexuality", label: "Sexuality" },
+  {
+    value: "place.stream.default.metadata#suffering",
+    label: "Upsetting or Disturbing",
+  },
+  { value: "place.stream.default.metadata#violence", label: "Violence" },
+];
```
espeon: I wonder if it would be better to get metadata from the lexicon. For stuff like this, if we change the lexicon we know due to type checks that this also needs to be changed 

Probably something like this would be nice, but also id be worried about rerenders with this specific implementation: 

```suggestion
const CONTENT_WARNINGS = (() => {
  // Find the metadata schema
  const metadataSchema = schemas.find(
    (schema) => schema.id === "place.stream.default.metadata",
  );

  if (!metadataSchema?.defs) {
    throw new Error("Could not find place.stream.default.metadata schema");
  }

  const contentWarningConstants = [
    { constant: PlaceStreamDefaultMetadata.DEATH, label: "Death" },
    { constant: PlaceStreamDefaultMetadata.DRUGUSE, label: "Drug Use" },
    {
      constant: PlaceStreamDefaultMetadata.FANTASYVIOLENCE,
      label: "Fantasy Violence",
    },
    {
      constant: PlaceStreamDefaultMetadata.FLASHINGLIGHTS,
      label: "Flashing Lights",
    },
    { constant: PlaceStreamDefaultMetadata.LANGUAGE, label: "Language" },
    { constant: PlaceStreamDefaultMetadata.NUDITY, label: "Nudity" },
    {
      constant: PlaceStreamDefaultMetadata.PII,
      label: "Personally Identifiable Information",
    },
    { constant: PlaceStreamDefaultMetadata.SEXUALITY, label: "Sexuality" },
    {
      constant: PlaceStreamDefaultMetadata.SUFFERING,
      label: "Upsetting or Disturbing",
    },
    { constant: PlaceStreamDefaultMetadata.VIOLENCE, label: "Violence" },
  ];

  return contentWarningConstants.map(({ constant, label }) => {
    // Extract the key from the constant by splitting on '#'
    const key = constant.split("#")[1];
    const def = metadataSchema.defs[key];
    const description = def?.description || `Description for ${label}`;

    return {
      value: constant,
      label: label,
      description: description,
    };
  });
})();
```

js/app/components/content-metadata-form.tsx
```diff
@@ -0,0 +1,1120 @@
...
+    "Attribution + non-commercial. Others may adapt and build upon your work for non-commercial purposes only, and must credit you.",
+  "place.stream.default.metadata#cc-by-nc-sa_4__0":
+    "Attribution + non-commercial + share-alike. Others may adapt and build upon your work for non-commercial purposes only, must credit you, and must license their new creations under identical terms.",
+  "place.stream.default.metadata#cc-by-nd_4__0":
+    "Attribution + no derivatives. Others may reuse your work, even commercially, but it must remain unchanged and you must be credited.",
+  "place.stream.default.metadata#cc-by-nc-nd_4__0":
+    "Attribution + non-commercial + no derivatives. Others may download and share your work with credit, but cannot change it or use it commercially.",
+  custom:
+    "Custom license. Define your own terms for how others can use, adapt, or share your content.",
+};
```
espeon: Same as above with these maybe?

js/app/components/content-metadata-form.tsx
```diff
@@ -0,0 +1,1120 @@
...
+
+// Helper to generate custom date options
+const generateDateOptions = () => {
```
espeon: Wonder if it's possible to generate the date array on the fly? For now this is probably good enough so long as dates are checked for validity on submit

js/app/components/content-metadata-form.tsx
```diff
@@ -0,0 +1,1120 @@
...
+                Content Warnings
+              </Paragraph>
+              <Paragraph color="$gray11" fontSize="$1">
+                (select any that apply)
+              </Paragraph>
+            </XStack>
+
+            {/* Ultra-compact grid layout for content warnings */}
+            <View flexDirection="row" flexWrap="wrap" gap={2} w="100%">
+              {CONTENT_WARNINGS.map((warning) => {
```
espeon, line 519: May want to add a info icon here with a popup with more details (or just description text if the screen is big enough)? Assuming people may want to know more about what exactly constitutes a specific content warning

js/app/features/bluesky/contentMetadataSlice.tsx
espeon, line 1: should possibly to move this over to a zustand store in @streamplace/components?

espeon (commented):
Will note that we're wanting to distribute some of this stuff (like the metadata form) in @streamplace/components, so nw on getting UI stuff with tamagui exactly right at the moment as it should all be moved over in time.

js/app/components/content-metadata-form.tsx
tripledoublev: date array is now dynamically generated and I added validation a7860395d692ff7d91433a9549797cfb2bf1db27

js/app/components/content-metadata-form.tsx
tripledoublev: Added description text for the selected content warnings
d0654853fdb9674779fa2fd2708a8da165766d1f
Not quite as you described (no pop up or icons) but it does provide description text to users when a warning is selected

js/app/components/content-metadata-form.tsx
tripledoublev: getting metadata from the lexicon is defninitely a better option. good suggestion
d3e5f88fe1d74ec17d25a875d835b5acfa2a6e02

js/app/components/content-metadata-form.tsx
tripledoublev: same as above
d3e5f88fe1d74ec17d25a875d835b5acfa2a6e02

lexicons/place/stream/default/metadata.json
uditvira: The one use I can think of is someone other than the streamer labelling it as PII after the fact. For example, a streamer (unintentionally or perhaps even maliciously) leaks PII into a stream, a content warning could serve some purpose. I guess a question would be whether this falls within the scope of the content warning feature to address or should be addressed by a separate mechanism with more teeth (i.e., moderation tools).

I am somewhat inclined to leave it in because taking it out would be a move that actively shifts us away from the standard, but I'm happy to follow your lead here. One possibility is to remove it as an option in the front end, but leave it as one of the known values everywhere else. What do you think?

js/app/components/content-metadata-form.tsx
espeon: ah this is fine! i was thinking popup b/c i thought it may be a lot of text

pkg/media/manifest_builder.go
uditvira: https://github.com/streamplace/streamplace/pull/467/commits/2361f468db4472d69196deeca7556184f6bdea6

js/app/features/bluesky/contentMetadataSlice.tsx
tripledoublev: is this possibly what you had in mind?
864c4bcad5d75f4837b0f17971a42d5a3d6c9555

pkg/media/manifest_builder.go
uditvira: https://github.com/streamplace/streamplace/pull/467/commits/5a970ffdcbcd9d2000e5f663ad4f54940a73d9de

iameli:
Closing in favor of https://github.com/streamplace/streamplace/pull/520

PR #481, opened 2025-08-18
build: dramatically simplify `make dev` cinematic universe

iameli:
Okay, getting the C dependencies working on macos consistently was always a pain, this simplifies some things: No more `streamplacedeps-unisntalled`. Instead, we do a `meson install` every time and have all of the actual dependencies in `./build-platform-arch/lib`. No more need to have `LD_LIBRARY_PATH` and `DYLD_LIBRARY_PATH` which I'm now convinced never really worked. The only prerequisite to running `go test` or `go build` on the command line is `PKG_CONFIG_PATH=$(realpath build-platform-arch/lib/pkgconfig)`.


iameli:
didn't work

PR #484, opened 2025-08-19
add irpc methods for stream coordination

b5:
The primary goal here is a viable strategy for near-term scalability: serving the same stream from multiple nodes, such that one node being overwhelmed can direct traffic to another replica instead of overwhelming the principal node.

This PR expands the existing `irpc` implementation with a new, basic set of methods for coordinating peers that exist, and what streams they are currently replicating. We briefly entertained using a gossip swarm for this, but building directly on irpc makes more sense in the short-term 1-to-500 node scale. 

### irpc messages added:
    - `peers`: returns a stream of all known peers & the topics they’re subscribed to
    - `announce_subscriptions`: send a message to another node, informing it of the streams your node is subscribed to
    - `get_subscriptions`: ask a peer for the list of streams it has (the “pull” version of `announce_streams`)

### TODO:
- [x] finish initial irpc implementation
  - [ ] maintain a list of streams the current node is subscribed to, return that in the `get_subscriptions`
- [x] require a configured set of bootstrap nodes for the `Receiver` struct
- [x] broadcast an `announce_streams` to all known peers when subscribing to a stream in the receiver
- [x] periodically call `list_streams` on bootstrap nodes to hydrate the peers list
- [x] broadcast `announce_streams` to all known peers once a minute
- [x] prune any node that hasn't sent an `announce_streams` in 2 minutes


iameli:
Closing in favor of https://github.com/streamplace/streamplace/pull/513

PR #490, opened 2025-08-20
feat: add canonical url in livestream record

espeon:
This would help certain individuals if one wanted to access one's livestream via e.g. bluesky statuses

### DIFF
--- a/js/components/src/hooks/useLivestreamInfo.ts
+++ b/js/components/src/hooks/useLivestreamInfo.ts
@@ -21,7 +21,10 @@ export function useLivestreamInfo(url?: string) {
       if (title !== "") {
         setRecordSubmitted(true);
         // Create the livestream record with title and custom url if available
-        await createStreamRecord(title, undefined, undefined, url);
+        await createStreamRecord({
+          title,
+          customUrl: url || null,
+        });
       }
     } catch (error) {
       console.error("Error creating livestream:", error);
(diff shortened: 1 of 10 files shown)


pkg/director/stream_session.go
```diff
@@ -304,7 +304,7 @@ func (ss *StreamSession) UpdateStatus(ctx context.Context, repoDID string) error
 		EmbedExternal: &bsky.EmbedExternal{
 			External: &bsky.EmbedExternal_External{
 				Title:       lsr.Title,
-				Uri:         fmt.Sprintf("https://%s/%s", ss.cli.PublicHost, repo.Handle),
+				Uri:         *lsr.CanonicalUrl,
```
iameli: This is optional and will panic if the record doesn't have one - gotta check earlier and use the previous thing instead

iameli (changes requested):

js/components/src/hooks/useLivestreamInfo.ts
```diff
@@ -20,7 +20,8 @@ export function useLivestreamInfo() {
     try {
       if (title !== "") {
         setRecordSubmitted(true);
-        await createStreamRecord(title);
+        // Create the livestream record with title and custom url if available
+        await createStreamRecord(title, undefined, undefined, url);
```
iameli: Kind of an awkward signature, an object type with optional parameters might be cleaner?

iameli (approved):
One suggested change but I know you're crunching, LGTM :)

js/components/src/hooks/useLivestreamInfo.ts
espeon: ooh good point, this is def doable

iameli (approved):

PR #495, opened 2025-08-24
player: pass object fit (and make package.json more lenient)

espeon:


js/components/package.json
```diff
@@ -30,28 +24,31 @@
...
     "viem": "^2.21.44",
-    "zustand": "^5.0.5"
+    "zustand": "^5.0.5",
+    "streamplace": "0.7.2"
   },
   "peerDependencies": {
     "react": "*"
+  },
+  "scripts": {
+    "build": "tsc",
```
iameli, line 51: oh yeah this nonsense again... we should at least get this here so we can't publish without building

```suggestion
    "build": "tsc",
    "prepublishOnly": "tsc",
```

iameli (approved):

iameli:
CI failing just for lockfile change reasons, needs `pnpm install` and commit the lockfile

js/components/package.json
espeon: hmm, seems to be ignoring place.stream.chat.gate? super wacky

iameli: wdym?

espeon: oh nvm found the issue, somehow set the streamplace dep in node_modules to something that's not the local one

PR #517, opened 2025-08-28
player: add play button if we can't autoplay on web

### LINKED ISSUE #514: Deny autoplaying video leaves you with no way to start the player
You can't start the player if you don't have autoplay on (which isn't uncommon afaik)

https://github.com/user-attachments/assets/1eecdac8-30e0-4897-bece-49255aa19af0

espeon:


iameli:
How do I test this? I've never had a muted play fail

espeon:
You can change 'autoplay' to 'block audio and video' and it should pop up
image

iameli (approved):
Tested successfully! (If you wanna go extra far we could drop a thumbnail under the video before it plays but this is a rare case, I'm not too worried about it)

espeon:
oh good point, may experiment tomorrow if i have time

PR #520, opened 2025-08-28
feature: add c2pa user consent

iameli:


iameli (commented):
Remaining issues on this:

* Move frontend from `app` to `components` and get rid of the tamagui/Redux elements, we're moving away from those
* Integrate with the new livestreaming dashboard
* Change the "expiry time" to a "delete my segments after N duration" time. That then gets translated into a expiry time when the segments themselves get minted. e.g. in the lexicon `expiry: 300` gets translated into an expiry time five minutes from now, so each segment gets deleted at that point

iameli:
Closed in favor of https://github.com/streamplace/streamplace/pull/656

PR #522, opened 2025-08-29
feat: temp error reporting

espeon:


iameli:
I asked you to do this but I think I've changed my mind, we're just gonna bite the bullet and add Sentry

PR #542, opened 2025-09-03
player: store global volume state

espeon:
this stores the latest volume/mute state in localstorage on web.


js/components/src/streamplace-store/streamplace-store.tsx
```diff
@@ -40,6 +46,46 @@ export const makeStreamplaceStore = ({
...
+      }
+
+      if (storedMuted) {
+        initialMuted = storedMuted === "true";
+      }
+    } catch (e) {
+      console.warn("Failed to load volume settings from storage:", e);
+    }
+  }
+
```
iameli: Might be time to port over the Storage class into `components`, yeah? That'll give us this and also work on native. https://github.com/streamplace/streamplace/tree/bdf804a95970207319c60a303ddb9ad1d7800a8d/js/app/storage

iameli (changes requested):
The volume parts all seem fine but seems like a good chance to move the Storage stuff over rather than checking for the existence of `localStorage` in two different places

Granted we don't have volume controls on native but it'd be nice to keep it all in one code path

PR #546, opened 2025-09-04
login: few small tweaks

espeon:
1. removed signup flow, always "sign up with Bluesky"
2. example placeholder
3. right-aligned all the buttons


js/app/components/login/login.tsx
```diff
@@ -141,6 +144,7 @@ export default function Login() {
               </Text>
               <Input
                 id="pdsUrl"
+                placeholder="alice.bsky.team"
```
iameli: not `bsky.social`? concerned someone might type in `someuser.bsky.team` in response to this
```suggestion
                placeholder="example.bsky.social"
```

iameli (approved):
placeholder change suggested but LGTM

js/app/components/login/login.tsx
espeon: oh good point haha. shyed away initially because alice.bsky.social (which most ppl use as a placeholder) is taken lol

PR #556, opened 2025-09-07
ui: force dark theme in the rest of the themeproviders

espeon:


iameli (approved):
Huh. I mean I'm cool with this but do we not just want one `<ThemeProvider>` way up in `app/components/provider/provider.shared.tsx`?

espeon:
...yeah good point haha

iameli (approved):

PR #572, opened 2025-09-15
feat: sentry

espeon:
Will listen to these config variables:

```sh
SP_ENABLE_SENTRY=true
SP_SENTRY_PROJECT=your-production-project
SP_SENTRY_ORG=your-production-org
EXPO_PUBLIC_SENTRY_DSN=your-production-dsn
```

Sentry in JS is always enabled, but should be silent if the DSN isn't set. [Spotlight](https://spotlightjs.com/about/) is enabled in dev environments.

### DIFF
--- a/js/app/components/provider/provider.shared.tsx
+++ b/js/app/components/provider/provider.shared.tsx
@@ -26,6 +26,7 @@ export default Sentry.wrap(ProviderInner);
 import * as Application from "expo-application";
 import Constants from "expo-constants";
 import * as Updates from "expo-updates";
+import { Platform } from "react-native";
 Sentry.setExtras({
   manifest: Updates.manifest,
   linkingUri: Constants.linkingUri,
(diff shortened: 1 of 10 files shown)


js/app/components/provider/provider.shared.tsx
```diff
@@ -20,13 +21,48 @@ import { store } from "store/store";
...
+
+function ProviderInner({
   children,
   linking,
 }: {
   children: React.ReactNode;
   linking: LinkingOptions<ReactNavigation.RootParamList>;
 }) {
+  Sentry.init({
+    dsn: process.env.EXPO_PUBLIC_SENTRY_DSN || undefined,
```
iameli: This is fine for iOS/Android - on web I'm not gonna do a special build for this, we run the same Streamplace as everybody else. So we're going to need a backend command-line argument for this key and we'll need to get it to the frontend somehow. One way would be to use the HTML-manipulation code that does OpenGraph linking to insert it into the page before the JS bundle loads? Probably more than one way to do it tho https://github.com/streamplace/streamplace/blob/next/pkg/linking/linking.go

iameli (changes requested):

js/app/components/provider/provider.shared.tsx
espeon: Should probably be good to go now? I'm a little worried that the DSN won't be tree-shaken on web (I just use Platform.OS to say if we're on web use the injected value), do you think that's a big enough issue?

iameli: Nah, we've got bigger fish to fry on reducing bundle size

iameli (approved):
LGTM!

PR #589, opened 2025-09-22
add atproto proxy header to app.bsky.actor.getProfile

espeon:
Without the atproto-proxy header, some pdses may not be able to auto-proxy bluesky appview requests, which means users get a 403 and can't log in. The app.bsky.actor.getProfile routes to the bsky appview and currently lacks this header, so we add it to enable proper proxying.

### DIFF
--- a/pkg/spxrpc/app_bsky_actor.go
+++ b/pkg/spxrpc/app_bsky_actor.go
@@ -16,6 +16,11 @@ func (s *Server) handleAppBskyActorGetProfile(ctx context.Context, actor string)
 		return nil, echo.NewHTTPError(http.StatusUnauthorized, "oauth session not found")
 	}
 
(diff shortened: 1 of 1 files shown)


iameli (approved):
Could you do a quick grep to check where else we're calling OATProxy's `client.SetHeaders`? I thiiiiiink it's only used in `com.atproto.moderation.createReport` to send it to the right labeler but I'm not 100% on that. I should probably add a `client.AddHeaders` function so we don't have to worry about clobbering things

But if we're not clobbering it anywhere go ahead and merge, LGTM

espeon:
looks exactly the same as the impl here (except capitalized differently, so will change that for consistency) so i'm assuming it's fine?

iameli:
Yep sounds good

PR #603, opened 2025-09-23
feat: i18n framework

### LINKED ISSUE #497: app: research best i18n solution

espeon:


js/docs/src/content/docs/guides/localizing/for-devs.md
```diff
@@ -0,0 +1,128 @@
...
+import { Trans as T } from "react-i18next";
+
+function HelpText() {
+  return (
+    <T
+      i18nKey="help-message"
+      components={{
+        supportLink: <a href="mailto:support@stream.place" />,
+        docsLink: <a href="/docs" />,
+        bold: <strong />,
```
iameli: This won't work in React Native, right?

iameli (changes requested):
Let's chat about the workflow here, want to make sure it's easy as possible for devs + translators here. Ideally it wouldn't require a manual step in between devs adding a new translation string and seeing it pop up on the app

Random stuff:

1. Language settings don't seem to persist across refresh
2. Feeling like this should probably be a `components` thing rather than an `app` thing, assuming our users are gonna want to provide translations as well
3. We've got several copies of everything here, ideally even if we need it copied we'd have one set of `.ftl` files committed and everything else would get generated
4. `scripts/extract-i18n.js` seems to clobber the organization in the .ftl files, is the plan to not use it?

js/docs/src/content/docs/guides/localizing/for-devs.md
espeon: yeah you're right it shouldn't. good point

espeon:
Fair for 1. My original idea was for apps to manage their own translations, but on reflection I don't think that would work as well as one would think.
2. Yeah, same. Above isn't going to pull through, so this makes best sense inside components.
3. Hopefully if we change 2, we should cut down on 3.
4. It should keep i18ns the same, and put all new keys in the json file. Key migration (the step after) puts all new keys in common.ftl. I think it would probably be prudent to combine those two scripts so it's just one thing to run haha.

js/components/src/i18n/i18next-config.ts
```diff
@@ -0,0 +1,275 @@
...
+            translations = require("../../public/locales/es-ES/messages.json");
+            break;
+          case "zh-Hant":
+            translations = require("../../public/locales/zh-Hant/messages.json");
+            break;
+          case "fr-FR":
+            translations = require("../../public/locales/fr-FR/messages.json");
+            break;
+          default:
+            throw new Error(`No static translation file for locale: ${locale}`);
```
iameli: I was about to tell you that this wouldn't work, 'cause the compiler isn't smart enough to not bundle these at build-time just because they're in the else block of `if (typeof window !== "undefined") {`. But... it totally is, they don't get bundled on web. Huh. Good job, Metro.

js/components/scripts/compile-translations.js
```diff
@@ -0,0 +1,210 @@
...
+    }
+  }
+
+  // Handle any remaining key
+  if (currentKey) {
+    translations[currentKey] = currentValue.join("\n").trim();
+  }
+
+  return JSON.stringify(translations, null, 2);
+}
```
iameli: Surely there's a library for this so we don't have to hand-roll it? Looking at https://github.com/projectfluent/fluent.js

iameli (changes requested):
Cool, getting there:
 
* Looks like the docs page isn't in the sidebar
* Suggested the use of a library for fluent.js stuff so we don't miss corner cases
* At least in dev, language selection seems to be not persistent; I'm getting English again after I refresh
* (Also in dev) it seems like I'm getting a FOUC on page load - settings page says `app-version` and whatnot for a moment until the translation loads. We can just do a `visibility: hidden` on everything until you get your localization, I reckon
* For a user adding translations, the dev flow doesn't totally work; you can run `i18n:watch` but that doesn't copy translations into `js/app/public` so you don't see the new translations in dev. It'd be nice if `i18n:watch` ran one command that built and copied the translations to where they need to be
* Need `nodemon` as a `devDependencies` in `components`, wasn't installed for me and it failed

I think there's room to improve the dev workflow further here but I don't wanna block getting this merged 👍

js/components/src/i18n/i18next-config.ts
espeon: I was surprised myself honestly

js/components/scripts/compile-translations.js
espeon: hmm, looks right: iNaturalist does basically that here: https://github.com/inaturalist/iNaturalistReactNative/blob/f6b3ba1841e9c1c93f5530a68c0fa67c622fcd05/src/i18n/i18ncli/ftl.js#L175

espeon:
Should have fixed all those issues, except the docs page in the sidebar, I couldn't replicate that and I don't think it should have been affected by this at all

iameli (approved):

espeon:
Will refactor the selection off and then commit

PR #604, opened 2025-09-24
ui: remove tamagui fixes 2

### LINKED ISSUE #592: tamagui removal cleanup
* Blank icon overlayed over profile icon in the corner
* Sidebar starts out too wide?

espeon:


js/app/components/mobile/player.tsx
```diff
@@ -55,6 +55,12 @@ export function Player(
 
   const navigation = useNavigation();
 
+  useEffect(() => {
+    return () => {
+      StatusBar.setHidden(false, "slide");
+    };
+  }, []);
+
```
iameli, line 63: Help me understand the difference between this and `<StatusBar hidden={true} />` below?

iameli (changes requested):
Just tested and I'm getting really poor performance on Android, very similar to before we memoized chat -- if you have any suspicions what could be causing that have a look but otherwise I can pull out the profiler again

js/app/components/mobile/player.tsx
espeon: All the logic maybe should be in that useEffect. They do two different things, the `<StatusBar hidden={true} />` hides the status bar constantly, and this excerpt shows it again on unload.

iameli: Oh I totally didn't clock that it was returning a value for cleanup instead of just calling it right away, LGTM

iameli (approved):
Tested again on Android and working fine, let's go!

PR #605, opened 2025-09-24
fix: proxy com.atproto.repo.* correctly 

### LINKED ISSUE #612: Color doesn't save
I set my color on the login page and it doesn't "stick" between reloads and/or another machine if I go login elsewhere

espeon:


pkg/spxrpc/com_atproto_repo.go
```diff
@@ -18,6 +19,23 @@ import (
 	"stream.place/streamplace/pkg/log"
 )
 
+func resolveRepoService(ctx context.Context, repo string) (did, service, handle string, err error) {
+	did = repo
```
iameli: I'm not really a fan of Go's named return types, I'm initially confused by it every time. Plus you're not even really using it, the bottom line could just say `return` -- let's just make the regular variables?

iameli (changes requested):
Nitpick change

PR #606, opened 2025-09-25
fix: set max lines for livestream title from 1 to 3

espeon:


iameli (approved):
I'm Sitting Here Trying To Think of a Hilarious Light Novel Title To Put Into This LGTM Message But Nothing is Coming To Mind?!?!?

PR #610, opened 2025-09-26
player: add manual rotation

espeon:
Implements rotation for the native video player via a rotation provider and accompanying hook.

Will need a native release, as we add `expo-screen-orientation`.

### DIFF
--- a/js/app/components/mobile/ui.tsx
+++ b/js/app/components/mobile/ui.tsx
@@ -10,6 +10,7 @@ import {
   useMuted,
   usePlayerDimensions,
   usePlayerStore,
+  useRotation,
   useSegmentDimensions,
   useSetMuted,
   useSetVolume,
@@ -23,6 +24,7 @@ import {
   ChevronRight,
   Fullscreen,
   Minimize,
+  RotateCw,
   SwitchCamera,
   Volume2,
   VolumeX,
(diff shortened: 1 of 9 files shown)


js/app/components/mobile/ui.tsx
```diff
@@ -375,6 +377,9 @@ function RightControlsPanel({
   const setMuted = useSetMuted();
   const fullscreen = usePlayerStore((x) => x.fullscreen);
   const setFullscreen = usePlayerStore((x) => x.setFullscreen);
+  const { toggleRotation, canRotate, currentOrientation } = useRotation();
+
+  console.log("rotation", canRotate, currentOrientation);
```
iameli:
```suggestion
```

js/components/src/components/mobile-player/rotation-lock.tsx
```diff
@@ -0,0 +1,162 @@
+import * as ScreenOrientation from "expo-screen-orientation";
```
iameli: Anything that's a new native dependency in @components we need to make optional, I think, we can't force our users to definitely 100% have this installed during an OTA update

I think all you gotta do is wrap this in a async loader (like video-async.native.tsx) and then use require() to check if the module is installed https://github.com/streamplace/streamplace/blob/next/js/components/src/crypto-polyfill.native.tsx

Actually you might have to figure out how to native module detection, as require() will just tell you if the JS is there

iameli (approved):
Optional thing

iameli (approved):
LGTM

PR #619, opened 2025-09-27
feat: make the Stream Key display less suck

BarryCarlyon:
# Reasoning:

1. Theres way too much padding on the fields in the Stream settings zone.

2. The first thing I did when using stream.place was leak my stream key as I was live on Twitch at the time. Protect the ID 10 T's.

3. This ads an explict copy to clip board button/function. Without the need to show the key first. (similar in function to Twitch's)

image

4. Remove forced to clip board of the key as thats bad User behaviour
5. Add Toast on clipboard yay/nay from streamplace/components
6. Tweak border raidus on RTMP/Whip button group

# Additional:

~~I'd be inclined to remove the "auto copy to clipboard" as this is a bad user UX. I didn't but I suggest we should and can add this to the PR.~~

I didn't also add a copy button to the Server URL but that could be added easily enough.
Both setting fields are easier to to copy from now anyway. (When I did my intial setup I err'ed and had a space at the start of the Server URI when added to OBS)

# Commit:

- Remove the slew of padding in this box
- Change the server and stream key display to readonly inputs Add an explict copy to clipboard button
- Change streamkey field to password and show it/select all on click
- Throw it into a ScrollView on the box so it scrolls on smoll height browsers and Tweak padding
- RTMP/Whip button group tidy
- Remove forced clipboard on key gen

# Screenshots

image
image

fix:
image
to:
image

### DIFF
--- a/js/app/components/live-dashboard/stream-key.tsx
+++ b/js/app/components/live-dashboard/stream-key.tsx
@@ -2,17 +2,18 @@ import { Body, Button, Code, Row, View } from "@streamplace/components";
 import { Redirect } from "components/aqlink";
 import Loading from "components/loading/loading";
 import {
-  clearStreamKeyRecord,
-  createStreamKeyRecord,
-  selectIsReady,
-  selectUserProfile,
+    clearStreamKeyRecord,
+    createStreamKeyRecord,
+    selectIsReady,
+    selectUserProfile,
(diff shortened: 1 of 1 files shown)


iameli (approved):
Looks sick, thank you so much!

PR #622, opened 2025-09-27
nit: Set vscode settings tabSize to 2

BarryCarlyon:
My vscode was resetting itself to 4 and messing up indenting. (my default is 4 but the project uses 2)

So set at project level for all vscode users


iameli:
thanks, LGTM!

PR #632, opened 2025-09-29
player: add follow button + link to bluesky

### LINKED ISSUE #486: no way to get to bluesky/stream.place profile from stream
i was expecting either the username or profile image to be clickable but that's not the case.

espeon:
should close #486 and reclose #102

Mobile:
image

Desktop: 
image

A few more changes were made re: typography, adding 'graph' manager hook

### DIFF
--- a/js/components/src/streamplace-store/graph.tsx
+++ b/js/components/src/streamplace-store/graph.tsx
@@ -1,0 +1,232 @@
+import { AppBskyGraphFollow } from "@atproto/api";
+import { useEffect, useState } from "react";
+import { useStreamplaceStore } from "./streamplace-store";
+import { usePDSAgent } from "./xrpc";
+
+export function useCreateFollowRecord() {
+  let agent = usePDSAgent();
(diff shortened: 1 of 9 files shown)


js/components/src/streamplace-store/graph.tsx
```diff
@@ -0,0 +1,232 @@
...
+      if (data.follow) {
+        setIsFollowing(true);
+        setFollowUri(data.follow.uri);
+      } else {
+        setIsFollowing(false);
+        setFollowUri(null);
+      }
+    } catch (err) {
+      setError(
+        `Could not determine follow state: ${err instanceof Error ? err.message : "Unknown error"}`,
```
iameli: In the event that we're catching a not-error it's almost certainly a string? Maybe 
```suggestion
        `Could not determine follow state: ${err instanceof Error ? err.message : `Unknown error: ${err}`}`,
```

iameli (approved):
LGTM, very minor point

PR #633, opened 2025-09-30
player: 'bullet comments'

espeon:
im really big on this unsurprisingly

Example (it scales to video size, which in this case is 16:9): 
image

With long messages:
image

I thought it would be fun to prototype something like this in my free time but it turned out so well I thought I should just commit it. We could also save this til april fools next year and give the site a niconicodouga reskin or something lol


espeon:
Will say I'm kind of unsure whether they should be distributed as part of the library. It's not really something most people would use in their own streamplace apps, and copying from our source is always an option there as well.

PSingletary:
NEAT

espeon:
demoted to easter egg

js/app/components/mobile/desktop-ui/mu.tsx
```diff
@@ -0,0 +1,28 @@
...
...45 Q 73.756 169.45 73.756 167.7 Q 73.756 165.95 74.756 164.95 Q 75.756 163.95 77.506 163.95 Z M 96.506 97.45 L 96.506 125.45 Q 96.506 127.45 98.006 128.95 Q 99.506 130.45 101.506 130.45 L 197.506 130.45 Q 199.506 130.45 201.006 128.95 Q 202.506 127.45 202.506 125.45 L 202.506 97.45 L 96.506 97.45 Z M 96.506 62.45 L 96.506 89.95 L 202.506 89.95 L 202.506 62.45 Q 202.506 60.45 201.006 58.95 Q 199.506 57.45 197.506 57.45 L 101.506 57.45 Q 99.506 57.45 98.006 58.95 Q 96.506 60.45 96.506 62.45 Z M 166.506 55.2 L 161.006 51.95 Q 168.756 41.7 176.881 27.7 A 383.321 383.321 0 0 0 189.09 5.114 A 264.812 264.812 0 0 0 190.506 2.2 Q 191.006 0.7 192.506 0.2 Q 194.006 -0.3 195.756 0.45 Q 197.256 0.95 197.756 2.45 Q 198.256 3.95 197.506 5.2 Q 190.506 17.95 182.506 31.45 Q 174.506 44.95 166.506 55.2 Z M 107.672 1.117 A 4.458 4.458 0 0 0 106.006 1.45 Q 104.756 1.95 104.381 2.95 Q 104.006 3.95 104.756 4.95 Q 111.756 13.95 117.006 22.825 Q 122.256 31.7 126.006 39.95 Q 126.506 41.2 127.756 41.7 Q 129.006 42.2 130.256 41.45 Q 132.006 40.7 132.506 39.2 Q 133.006 37.7 132.256 36.45 Q 128.506 28.45 123.256 19.95 Q 118.006 11.45 110.756 2.45 Q 109.756 1.45 108.506 1.2 A 4.24 4.24 0 0 0 107.672 1.117 Z" />
```
iameli, line 23: haha probably if there was more than the one path it'd make sense to split these into constants for readability but NBD for one. just kinda funny.

iameli (approved):
Totally shippable as-is and I like the easter egg concept in general but let's figure out how to turn this into an actual feature at some point for sure, I love stuff like this for giving us some flavor. Thoughts:

1. Could ship an embeddable overlay so streamers can burn the bullets into the stream if they want to
2. We can also totally make it streamer opt-in as part of the branding work

Having this be a streamer-customization thing rather than a client-specific per-app viewer option seems more coherent to me (although we could retain "turn on danmu for all streams" as a secret easter egg option I suppose)

espeon:
yeah, doing either (or both!) is totally doable! will do a rebase then that first thing then will merge ! think

PR #639, opened 2025-09-30
integrate iroh-kv into streamplace

iameli:
Got this broadly working! Right now it's just every node publishing `Put(nil, streamerName, segmentRecieved)` but that's a foundation we can build a lot on.

Some notes as I did this:

1. ~~I dunno if this is a bug with the new uniffi-bindgen-go version, but it appears to have problems with Rust functions that return a `Result` that will not error under any circumstances. I guess because there's no error to work with...? Anyway I had to do some stuff like this to make the functions not panic:~~

**Update: Never mind, it was just https://github.com/NordSecurity/uniffi-bindgen-go/issues/79. Switched to the fork from https://github.com/NordSecurity/uniffi-bindgen-go/pull/82 and we're good to go here.**

2. I need some kind of mechanism for cleanly shutting the server down, how should I go about that? (On the Go side, this will look like our context being cancelled, see the `ctx.Err()` check in the subscribe loop.)
3. `sub := db.Subscribe(iroh_streamplace.NewFilter())` eventually just starts returning `nil` every time, is that intentional? Is there a mechanism for me to block forever, waiting on new information? I had to put it in a loop with a `time.Sleep(1 * time.Second) to prevent it from being an infinite loop.
4. I implemented storage of the node's secret key so every time it boots up it has the same one. I noticed that the text of the `ticket` that I give to other nodes seems to change every time. Despite that, after restarting the node, it seems to come up just fine with an old ticket. Is this the right mechanism for new nodes joining the cluster -- get a ticket from Node A and provide it to all the other nodes?


rklaehn (approved):
I see you already removed the old stuff. I can move things around a little bit then - should not affect the view from golang.

Also, you removed my test code. So how do I test the thing now with minimal hassle?

rklaehn:
> sub := db.Subscribe(iroh_streamplace.NewFilter()) eventually just starts returning nil every time, is that intentional? Is there a mechanism for me to block forever, waiting on new information? I had to put it in a loop with a `time.Sleep(1 * time.Second) to prevent it from being an infinite loop.

It is not. It should block until there is a new msg. If it does not, that's a bug or something about uniffi I don't understand. I think NewFilter should give you a filter that matches everything.

> I implemented storage of the node's secret key so every time it boots up it has the same one. I noticed that the text of the ticket that I give to other nodes seems to change every time. Despite that, after restarting the node, it seems to come up just fine with an old ticket. Is this the right mechanism for new nodes joining the cluster -- get a ticket from Node A and provide it to all the other nodes?

The ticket will contain information which rarely changes, such as the relay URL that does not change unless you move to a different region of the world, and information that changes frequently, such as the IP addr (changes e.g. if you move to a different wifi) and the randomly assigned port number.

Even without any explicit discovery mechanism, iroh will be able to dial a ticket as long as the relay URL is still correct. It will just try to dial the old ip addrs and give up quickly, so they don't do no harm.

If you have discovery configured, e.g. discovery_n0(), you don't even need the relay URL, since iroh will ask the discovery service(s) for the relay url for the node id.

At the very minimum a node ticket is just a node id encoded as text in a stable way.

PR #640, opened 2025-10-01
fix: Add problems back to live-dashboard

### LINKED ISSUE #599: Re-introduce Problems into the Live Dashboard
It's good that the Problems were hidden for the moment but with https://github.com/streamplace/streamplace/pull/597 landing we once again have solid B-Frame detection and can surface that to our users in the new dashboard.

espeon:
Note: I changed the design a tad to fit the new font, etc.

image

### DIFF
--- a/js/app/components/live-dashboard/bento-grid.tsx
+++ b/js/app/components/live-dashboard/bento-grid.tsx
@@ -9,6 +9,7 @@ import {
   useSegmentTiming,
   zero,
 } from "@streamplace/components";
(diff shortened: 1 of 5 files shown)


iameli (changes requested):
Looks great! Only problem problem that I'm finding is that the "long segments" detection doesn't kick in until you've got <3 segments and it doesn't seem to be popping up/maybe I dismissed it already? I think this is fixable by just adding a "X Problems Detected" button next to the stream health indicator so the user can see that and click it and see the problems come back. Totally shippable with that 👍

PR #641, opened 2025-10-01
fix: add proper web link behavior in sidebar

### LINKED ISSUE #634: Navbar links are not anchors
(web app v0.7.33)
The links/icons in the navigation drawer seem to navigate by an onclick event and don't contain anchor elements. This stops the browser from showing a right click menu ("open in new tab" etc.) and could produce a worse ux with accessibility tools.

espeon:
Should close #634, there should be proper anchors in sidebar links

image

### DIFF
--- a/js/app/components/sidebar/sidebar-item.tsx
+++ b/js/app/components/sidebar/sidebar-item.tsx
@@ -73,7 +73,13 @@ export default function SidebarItem({
     }
 
     // Fallback
-    console.log("tried to render item, but couldn't", (icon as any).$$typeof);
+    console.log(
+      "tried to render item for route",
+      label,
+      href,
+      ", but couldn't",
+      (icon as any).$$typeof,
+    );
 
     return <Text>📄</Text>;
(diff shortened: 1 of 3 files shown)


js/app/components/sidebar/sidebar-item.tsx
```diff
@@ -1,19 +1,25 @@
...
+  GestureResponderEvent,
+  Pressable,
   PressableStateCallbackType,
   StyleProp,
   View,
   ViewStyle,
 } from "react-native";
-import { Pressable } from "react-native-gesture-handler";
+
+import {} from "src/router";
```
iameli: ?

iameli (approved):
Looks fine aside from probably-accidental import?

js/app/components/sidebar/sidebar-item.tsx
espeon: Ah whoops, surprising that persisted 🤔

PR #642, opened 2025-10-02
feat: DoH checking in webhooks to avoid invalid/local ips

### LINKED ISSUE #595: Validate we're not hitting LAN IPs for webhooks and multistream targets

espeon:


pkg/aqhttp/aqhttp.go
```diff
@@ -17,5 +20,11 @@ func (adt *AddHeaderTransport) RoundTrip(req *http.Request) (*http.Response, err
 func init() {
 	Client = http.Client{
 		Transport: &AddHeaderTransport{T: &http.Transport{}},
+		// do not follow redirects automatically
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
```
iameli: Hmmmm nervous about what we're not thinking about that this might cause to fail. I guess we DO we want this behavior for handle and DID resolution, redirects are invalid there...

pkg/aqhttp/resolv.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	ranges := append(ipv4Bogons, ipv6Bogons...)
+	var invalidRanges []*net.IPNet
+	for _, cidr := range ranges {
+		_, network, err := net.ParseCIDR(cidr)
+		if err == nil {
+			invalidRanges = append(invalidRanges, network)
+		}
+	}
```
iameli, line 63: If this is just parsing static CIDRs, probably we should panic if this doesn't work? It'd imply there's an error in our static list of CIDRs

pkg/integrations/discord/send-chat.go
```diff
@@ -54,11 +61,24 @@ func SendChat(ctx context.Context, w *discordtypes.Webhook, did string, scm *str
...
+		// Use default port based on scheme
+		if parsedURL.Scheme == "https" {
+			port = "443"
+		} else {
+			port = "80"
+		}
+	}
+	requestURL := fmt.Sprintf("%s://%s%s", parsedURL.Scheme, net.JoinHostPort(targetIP, port), parsedURL.RequestURI())
+
+	req, err := http.NewRequestWithContext(ctx, "POST", requestURL, bytes.NewReader(jsonPayload))
```
iameli: Concerned about HTTPS here - a request to https://resolved_ip_address will fail on the TLS layer by default I think because the IP isn't in a SAN of the cert? Might need to go deeper on the TLS connection instantiation.

Or it just works but I'd be surprised, try it with a https://webhook.site URL

iameli (changes requested):
HTTPS concerns

iameli:
@thisismissem.social pointed this out too, possibly useful to use internally https://github.com/stripe/smokescreen

espeon:
seems like go on non-unix doesn't have `syscall.SIGUSR2` (fair on go's part, as sigusr2 isn't valid on windows). Could either 
1. create a generic version that isn't as protected on windows and have simpler (previous?) fetching behaviour
2. ditch smokescreen entirely
3. fork/patch smokescreen and add a case for windows

iameli:
@espeon Yeah this looks totally fine without Smokescreen, you can just ditch it.

We may need some kind of flag or option to disallow this behavior at some point or allow-list LAN IPs? Not for webhooks, but for other node-to-node requests. But we can ship this ahead of that.

espeon:
cool, will reverse those commits then 🫡

espeon:
if anyone is curious the branch with smokescreen is here: https://github.com/streamplace/streamplace/tree/natb/discord-no-lan-ips-with-smokescreen

iameli (changes requested):
Hah, you've done your job too well, our dev env failed:

```
2025/10/27 20:07:17 INFO starting firehose consumer relayHost=ws://localhost:33181 func=StartFirehose caller=pkg/atproto/firehose.go:122
    devenv.go:108: 
        	Error Trace:	/builds/streamplace/streamplace/pkg/devenv/devenv.go:108
        	            				/builds/streamplace/streamplace/pkg/atproto/chat_message_test.go:57
        	Error:      	Received unexpected error:
        	            	request failed: Post "http://localhost:33181/xrpc/com.atproto.server.createAccount": all resolved IPs for localhost are private/invalid
        	Test:       	TestChatMessage
    devenv.go:71: killing dev env
```

So we just need a simple way for us to separate out requests that should be restricted to public IPs and requests that shouldn't. Perhaps two exports from `aqhttp`? `aqhttp.Client` and `aqhttp.DangerousClient` or something? And then the devenv stuff can use the dangerous one.

espeon:
right now I have aqhttp.Do as the protected one and aqhttp.DoTrusted as unprotected so will switch the tests over haha

iameli:
Did a quick rebase, will merge if it passes

PR #644, opened 2025-10-04
feat: dialogs are sheets now actually 

espeon:
but only on mobile.

Also added proper sheet <> input interactions to text inputs, so the sheets focus properly on mobile as well.

| Non-Focused | Focused |
|--------------|----------|
| IMG_0244 | IMG_0245 |


iameli (approved):
Tested on Android, looks great!

PR #646, opened 2025-10-04
fix: upsert livestrem record

espeon:
Just inserting the livestream record seems to break our 'update livestream' feature. Even after we move over to more granular notification control, this is still useful in case you make a mistake or something in your livestream title.


iameli (approved):
sick

PR #651, opened 2025-10-07
fix: buttons with invalid/missing routes don't redirect home

### LINKED ISSUE #647: clicking on "documentation" redirects you back to home

espeon:


js/app/components/sidebar/sidebar-item.tsx
```diff
@@ -91,7 +91,7 @@ export default function SidebarItem({
       role="link"
       accessibilityLabel={typeof label === "string" ? label : "Link to " + href}
       // @ts-ignore This makes it render as <a> on web!
-      href={href}
+      href={route && href}
```
iameli: This'll be `href={false}` if it doesn't exist which is a little weird. Maybe:
```suggestion
      href={route ? href : undefined}
```

iameli (changes requested):
small thing

PR #652, opened 2025-10-08
feat: add stereo audio on ios + android builds

espeon:


iameli (approved):
Looks fine to me but I won't have a chance to test it properly until I'm back home on Saturday most likely (at which point I'll do a new native build with everything)

PSingletary:
will i finally have bluetooth out on iOS?

iameli:
@PSingletary I assume we won't have fixed that yet but it's maybe worth investigating if there's anything else we should be calling to that happen

espeon:
For what it's worth, my airpods work perfectly fine but they're in mono (assuming because it might activate communication mode or something?) Either way it's kind of out of my depth at this point haha

PR #653, opened 2025-10-08
Flatten the rust code structure

rklaehn:
This flattens the code structure of the rust part a bit.

It is a bit more flat than you would typically do in a rust crate, but since the golang symbols are exported at top level anyway, this matches more how the go bindings will look. You will get a rust compile error if you have 2 things with the same name instead of getting a weird uniffi error.


iameli (approved):
Yep LGTM!

PR #654, opened 2025-10-08
fix: ogimage dimension ration is actually spec now

espeon:
Set ratio to 1.9:1 as recommended by Facebook et al.

Also, fixed:
 - font reduction not working in og images
 - colors based on profile (pink default)

image


iameli (changes requested):
We don't gotta default to pink, we can pull their default color from `pkg/model/chat_message_default_colors.go`, you might have to reverse-engineer or refactor out my algorithm for deterministically picking one tho

EDIT: Alternately you could knock out https://github.com/streamplace/streamplace/issues/658 and then it doesn't really matter because you won't hit the default unless you're doing something weird

iameli:
Test failure is a known flake, LGTM

PR #655, opened 2025-10-09
feature: c2pa user consent 

tripledoublev:
follow up to https://github.com/streamplace/streamplace/pull/520 :

- moves content metadata UI form from `app` to `components` 
- integrates the form in the stream settings section of the livestreaming dashboard
- changes distribution policy expiry type from string to number (duration in seconds)

as requested in this [comment](https://github.com/streamplace/streamplace/pull/520#pullrequestreview-3166819091)

PR #656, opened 2025-10-09
feature: c2pa - metadata

tripledoublev:
#655 recreated on origin.


iameli:
This is looking really really great! I have a couple thoughts and I think it's good to ship:

I've been thinking about the problem of mapping Streamplace lexicon-defined metadata onto the C2PA segments themselves. Here's what I'm thinking: we're probably going to want the actual `place.stream.metadata.configuration` record embedded in the segments, while including other metadata record on a best-effort basis. That way there's no ambiguity mapping to and from somebody else's metadata standards for defining how the segments are handled; other Streamplace nodes  just look at the actual record embedded in the segment and be 100% confident of what the user is expressing. But we'll also do our best to translate our metadata definition into other people's metadata standards, purely to try and maximize interoperability with other tooling.

pkg/media/segmenter.go
```diff
@@ -91,6 +91,8 @@ func (mm *MediaManager) SegmentAndSignElem(ctx context.Context, ms MediaSigner)
 				if err != nil {
 					log.Error(ctx, "error validating segment", "error", err)
 					globalerror.GlobalError(err)
+					// Stop the pipeline to end the stream (removes from feed)
+					elem.ErrorMessage(gst.DomainCore, gst.CoreErrorFailed, err.Error(), fmt.Sprintf("Segment validation failed: %s", err.Error()))
```
iameli: I get why this _seems_ correct but we don't actually want to shut down the segmentation pipeline upon getting an invalid segment. We're ingesting WebRTC streams from phones with weird connections; in the event they send us some weird data (like an audio-only segment, which we don't presently support) we don't wanna bail out and force WebRTC renegotiation and stuff

PR #684, opened 2025-10-23
fix: make all app.bsky proxied requests unauthed

espeon:
Attempt to fix "signature doesn't match issuer" by making unauthed requests?


iameli:
Still useful or nah?

espeon:
Was just a shot in the dark honestly, I don't think it's necessary anymore

PR #685, opened 2025-10-23
golive: add bluesky push checkboxes, remove edit

iameli:
@espeon Please review. Last thought: maybe `url` and `canonicalUrl` should be swapped... it seems like url is the one people would show by default if they weren't thinking. Uncertain though.

image

### DIFF
--- a/js/app/components/live-dashboard/livestream-panel.tsx
+++ b/js/app/components/live-dashboard/livestream-panel.tsx
@@ -1,15 +1,19 @@
 import {
   Button,
   ContentMetadataForm,
+  Input,
   Textarea,
   useCreateStreamRecord,
   useLivestream,
   useToast,
   useUpdateStreamRecord,
+  useUrl,
   zero,
 } from "@streamplace/components";
+import { Checkbox } from "@streamplace/components/src/components/ui/checkbox";
+import { Tooltip } from "@streamplace/components/src/components/ui/tooltip";
 import { ImagePlus, X } from "lucide-react-native";
-import { useCallback, useMemo, useState } from "react";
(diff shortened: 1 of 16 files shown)


js/app/components/live-dashboard/livestream-panel.tsx
```diff
@@ -1,15 +1,19 @@
...
   Textarea,
   useCreateStreamRecord,
   useLivestream,
   useToast,
   useUpdateStreamRecord,
+  useUrl,
   zero,
 } from "@streamplace/components";
+import { Checkbox } from "@streamplace/components/src/components/ui/checkbox";
+import { Tooltip } from "@streamplace/components/src/components/ui/tooltip";
```
espeon: nit: Should probably be imported right from @streamplace/components?

espeon (approved):
Seems solid to me! The inactive color of the checkbox irks me so I'll change that (in another PR prob) but past that there's no bad issues that I can see UI-wise.

espeon:
afaik I added the canonicalURL specifically for skylight reasons? b/c there can be instances where one (e.g. skylight) may provide a "better"/more integrated viewing experience on their own app vs ours?

I'm not 100% on exposing it to the user to directly myself though, but there can be places where it could be neat? e.g. kidnotkin has a site (kidnotkin.io) he could put in there to direct users over that way?

iameli:
> I'm not 100% on exposing it to the user to directly myself though, but there can be places where it could be neat? e.g. kidnotkin has a site (kidnotkin.io) he could put in there to direct users over that way?

I did it for myself - I want to start streaming into my own node at iame.li, but I don't have the bandwidth to support lots of users so I want people sent over to the syndicated one

PR #686, opened 2025-10-23
fix local ref in lexicon files

notjuliet:
As specified in the specs, local ref values should be referenced starting with the hash part: https://atproto.com/specs/lexicon#ref
> The ref string can be a global reference to a Lexicon type definition (an NSID, optionally with a #-delimited name indicating a definition other than main), or can indicate a local definition within the same Lexicon file (a # followed by a name).


iameli:
Cool! Sounds reasonable, I just ran `make lexicons` to regenerate the docs, will merge if CI passes

PR #702, opened 2025-10-29
fix: clean up dropdown rendering a bit

### LINKED ISSUE #693: Stream Settings visibility issue
On iOS 26, when watching a video full screen, the settings overflow:

image

espeon:
wrapped some components to detect navigation stack context and render as simple `Pressable`/`View` components if we don't have said context.

also, show bottom sheet view in 'mobile mode' on the web where we don't show  the sidebar


iameli (approved):
I'm cool with this though at a certain point it'll be cleaner to split out a `dropdown.native.tsx` instead of having these big if blocks in the render functions. But this is totally mergable

espeon:
yeah actually I should just do that now to get it over with

PR #711, opened 2025-11-02
feat: add viewer-count component

justinmakaila:
Refactor viewers component to utilize a state-controlled viewer-count; can be utilized by external players.

I didn't want to utilize the whole component in my little player.

Want to support clients being able to change icons, etc, but don't want to do a complete refactor and start changing things without making small refactors to have the discussions.

Cheers!


espeon (approved):
lgtm, will check locally (i rebased to latest) but i don't see any issues

justinmakaila (commented):
I'm missing the leading prop. Will add that back.

PR #712, opened 2025-11-04
feat: branding system

### LINKED ISSUE #528: whitelabel: move all streamplace branding into a single, replaceable folder, make available through a useBranding() hook
1. Consolidate all logos and titles and stuff into a single folder so we have consistent keys to reference them
2. New StateDB table: `branding` with (`broadcasterId`, `key`, `mimeType`, `data`) (broadcasterId is how we're implementing multi tenancy)
3. On the backend for each of those keys we can do a lookup; either pull the custom logo/text out of the database or fall back to hardcoded defaults
4. XRPC endpoint to get this stuff, something like `place.stream.branding.getBlob?key=mainLogo`
5. For using this stuff, a new React hook: `useGetBranding()` that pulls stuff down based on the broadcaster context
6. Build a frontend for node admins to change all of those fields into whatever they want

There can also be a few other settings exposed through this interface, such as a "default stream" for single-user nodes (like iame.li) or customization of links on the sidebar

espeon:
Adds a scoped key(s)-value database to statedb to store branding assets in, and uses said assets to provide hooks to modify the app with.

## Features
- BrandingBlob model, connected to the statedb
- XRPC endpoints (getBranding, getBlob) with caching on client + server
- Simple (temporary? i hope so)Admin interface
- Hooks
- Single user mode

## supported keys

- `mainLogo`, `favicon` - images (SVG/PNG/JPEG/ICO)
- `siteTitle`, `siteDescription` - text metadata
- `primaryColor`, `accentColor` - theme colors
- `defaultStreamer` - optional auto-load DID

## technical notes

- soft delete handling with `Unscoped()` to avoid unique constraint violations on re-creation
- unauthenticated access for public branding
- multi-broadcaster supported via DID parameter
- single-user mode replaces the home page directly with the stream component in the router


js/.DS_Store
iameli, line 1: rip

```bash
echo '.DS_Store' >> ~/.gitignore_global && config --global core.excludesfile ~/.gitignore_global
```

js/app/src/screens/mobile-stream.tsx
```diff
@@ -1,32 +1,93 @@
...
+}
+
 export default function MobileStream({ route }) {
-  const { user, protocol, url } = route.params;
+  let user: string | undefined = route.params?.user;
+  const url: string | undefined = route.params?.url;
+
+  const defaultStreamer = useDefaultStreamer();
+
+  console.log(user, defaultStreamer);
```
iameli:
```suggestion
```

js/.DS_Store
espeon: i don't know how that got in there, thanks for spotting it 🥲

.DS_Store
iameli, line 1: hah

pkg/api/branding-admin.html
iameli, line 1: this can be gone now yeah? cool

pkg/spxrpc/place_stream_branding.go
```diff
@@ -0,0 +1,279 @@
...
+	"gorm.io/gorm"
+	"stream.place/streamplace/pkg/log"
+	placestreamtypes "stream.place/streamplace/pkg/streamplace"
+)
+
+//go:embed assets/logo.svg
+var defaultLogoSVG []byte
+
+//go:embed assets/favicon.svg
+var defaultFaviconSVG []byte
```
iameli: Can get these from the embedded application blob probably

pkg/spxrpc/place_stream_branding.go
```diff
@@ -0,0 +1,279 @@
...
+	data []byte
+	mime string
+}{
+	"mainLogo":         {data: defaultLogoSVG, mime: "image/svg+xml"},
+	"favicon":          {data: defaultFaviconSVG, mime: "image/svg+xml"},
+	"siteTitle":        {data: []byte("Streamplace"), mime: "text/plain"},
+	"siteDescription":  {data: []byte("Live streaming platform"), mime: "text/plain"},
+	"primaryColor":     {data: []byte("#6366f1"), mime: "text/plain"},
+	"accentColor":      {data: []byte("#8b5cf6"), mime: "text/plain"},
+	"defaultStreamKey": {data: []byte(""), mime: "text/plain"},
```
iameli: don't need i think

pkg/spxrpc/place_stream_branding.go
```diff
@@ -0,0 +1,279 @@
...
+
+	err = s.statefulDB.PutBrandingBlob(broadcasterID, input.Key, input.MimeType, data, width, height)
+	if err != nil {
+		log.Error(ctx, "failed to store branding blob", "err", err)
+		return nil, echo.NewHTTPError(http.StatusInternalServerError, "unable to store branding blob")
+	}
+
+	// invalidate cache
+	cacheKey := fmt.Sprintf("%s:%s", broadcasterID, input.Key)
+	s.BrandingCache.Delete(cacheKey)
```
iameli: This part won't work multi-node. I wouldn't sweat a server-side cache at all honestly; there's a client-side cache for end users and the database is fast

pkg/spxrpc/spxrpc.go
```diff
@@ -24,25 +24,27 @@ import (
...
-		model:        model,
-		OGImageCache: cache.New(5*time.Minute, 10*time.Minute), // 5min TTL, 10min cleanup
-		ATSync:       atsync,
-		statefulDB:   statefulDB,
-		bus:          bus,
+		e:             e,
+		cli:           cli,
+		model:         model,
+		OGImageCache:  cache.New(5*time.Minute, 10*time.Minute), // 5min TTL, 10min cleanup
+		BrandingCache: cache.New(1*time.Hour, 15*time.Minute),   // 1hr TTL, 15min cleanup
```
iameli: yeah i think we can lose it

pkg/api/api_internal.go
```diff
@@ -460,6 +460,44 @@ func (a *StreamplaceAPI) InternalHandler(ctx context.Context) (http.Handler, err
...
+		reader, err := a.XRPCServer.HandlePlaceStreamBrandingGetBlobDirect(ctx, broadcasterDID, key)
+		if err != nil {
+			errors.WriteHTTPInternalServerError(w, "failed to fetch blob", err)
+			return
+		}
+		if _, err := io.Copy(w, reader); err != nil {
+			log.Error(ctx, "error writing response", "error", err)
+		}
+	})
+
```
iameli: Gone also

js/app/components/mobile/user-offline.tsx
```diff
@@ -0,0 +1,218 @@
+import { useNavigation } from "@react-navigation/native";
```
iameli, line 1: Unintentional to be in this PR?

iameli (changes requested):
Great start! Let's do this to ship it:

- [ ]  Let's do some kind of XRPC server endpoint for "getBroadcasterSettings" that returns the list of admins; we can then use that to decide whether "branding" shows up on the settings page (we can just do this after the settings redesign ships)
- [x]  Ditch the server-side cache and figure out the client-side one – it'll be good also to have immediate feedback for the customizer when we figure that out.
- [x]  Default logo seems to be gone if you don't have one set
- [x]  Server-side favicon will be good

espeon:
fwiw: built the list of admins into getBroadcaster as it already returns node/broadcaster-specific configuration

iameli (approved):
LFGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGG

PR #713, opened 2025-11-04
fix: viewer # misalignment + home screen ratio

espeon:
Right now the first column looks like this - it's overly tall because the ratio is off.
image

On mobile, the number on the viewer count is misaligned (the line height is too high) which this fixes as well.


iameli (approved):
`DehydratedViewers` loooool

js/app/components/home/cards.tsx
```diff
@@ -101,7 +100,7 @@ const StreamCard = ({
               },
             ]}
           >
-            <Viewers viewers={viewers} />
+            <PlayerUI.DehydratedViewers viewers={viewers} />
```
iameli, line 103: lmao

iameli (approved):

PR #715, opened 2025-11-04
Do not use await in actor handler

rklaehn:
Previously we used await in the actor handler when subscribing and unsubscribing. This is less than optimal since the actor is blocked until the await completes.

This moves the handling into separate tasks.


iameli:
Closing in favor of https://github.com/streamplace/streamplace/pull/770

PR #716, opened 2025-11-05
Make C2PA manifest more idiomatic

makew0rld:
I read through the latest C2PA spec (2.2) and tried to align Streamplace with that.

Changes:
- Basic actions (`c2pa.created`, `c2pa.published`) are now timestamped
- The more standard [`cawg.metadata`](https://cawg.io/metadata/1.2-draft+media-identifiers/) is used over `place.stream.metadata`
- Streamplace metadata uses a namespace
- Removed unnecessary arrays from Dublin Core (`dc:`) metadata 

You'll note I made up a namespace URL for the JSON-LD, `https://ns.stream.place/metadata/0.1`. Let me know what you think about that. It doesn't need to resolve, although ideally it would. I left the version as `0.1` since things are bound to change, but ideally you would version metadata changes once things are stable.

The CAWG has some other assertions that might be useful, for example an [Identity](https://cawg.io/identity/1.2-draft/) assertion that seems to allow DID signing of specific C2PA claims by an individual. Perhaps something to explore in the future.


pkg/media/manifest_builder.go
```diff
@@ -242,7 +251,7 @@ func (mb *ManifestBuilder) enhanceManifestWithMetadata(mani obj, metadata *strea
 			// Note: In the manifest, we store this in "deleteAfter" field but with timestamp value instead of duration
 			deleteAfterTimestamp := aqtime.FromMillis(expiresAtSeconds * 1000).String()
 
-			mani["assertions"].([]obj)[1]["data"].(obj)["distributionPolicy"] = obj{
+			mani["assertions"].([]obj)[1]["data"].(obj)["streamplace:distributionPolicy"] = obj{
```
iameli: Now that we've made the decision to have the embedded Streamplace atproto records be the source of truth for how it gets distributed, I'm wondering if there's any point in having this field at all. If you can find any interoperable spec for defining distribution preferences I'd be happy to include that but if it's Streamplace-specific maybe now there's no point. I looked through the IPTC stuff without finding anything that looked like a "delete after" field (which surprised me!)

js/docs/src/content/docs/video-metadata/c2pa-integration.md
```diff
@@ -83,26 +85,30 @@ manifest, extracted from the MP4 segment using
...
-          "label": "place.stream.metadata",
+          "label": "cawg.metadata",
           "data": {
             "@context": {
+              "Iptc4xmpExt": "http://iptc.org/std/Iptc4xmpExt/2008-02-29/",
               "photoshop": "http://ns.adobe.com/photoshop/1.0/",
               "dc": "http://purl.org/dc/elements/1.1/",
               "xmpRights": "http://ns.adobe.com/xap/1.0/rights/",
-              "Iptc4xmpExt": "http://iptc.org/std/Iptc4xmpExt/2008-02-29/"
+              "streamplace": "https://ns.stream.place/metadata/0.1"
```
iameli: I think there's some reason they mostly use `http://`, could you look into it? I remember reading about this for a while

iameli (changes requested):
Looks like tests are failing. Past that, one potential simplification on the metadata

js/docs/src/content/docs/video-metadata/c2pa-integration.md
makew0rld: Some quick searches seemed to indicate http is historical and https is best practice.
- https://github.com/schemaorg/schemaorg/issues/2886
- https://schema.org/docs/developers.html#:~:text=Although%20we,change%20this

Not relevant if this is all getting removed however.

iameli: Gotcha. I'm not against publishing a JSON-LD schema if it's useful but yeah let's just not use it if we don't need it

makew0rld:
I've done what I can regarding test failures, the remaining ones seem to be GStreamer leaks and not related to my code changes? It's a bit unclear to me. Also it seems I had to add `-lm` to get `make dev-test` to work. Let me know if that was in error.

makew0rld:
Redundant distributionPolicy field has been removed as requested.

makew0rld:
Closed in favour of #775

PR #732, opened 2025-11-20
fix: untrusted resolution blocking all v4 addresses

espeon:
it seems like all v4 addresses were blocked. Have included tests that makes sure we can resolve to public v4s, and check that we throw an error on localhost/bogons resolution with the untrusted client.


iameli (approved):
cool!

PR #735, opened 2025-11-21
fix: make rotation layout adjustment slightly quicker?

espeon:
- Basically: use SafeAreaView (instant) for offsets instead of using useSafeAreaInsets (not instant).


iameli (approved):
LGTM AFAICT

PR #737, opened 2025-11-22
feat: split login out to a modal

### LINKED ISSUE #736: Redirect back to page that caused the login
Thinking the play is to make the login page an overlay with the same design, and on login we call a function that shows that page?

Image

https://bsky.app/profile/hipstersmoothie.com/post/3m65w23acrc2z

espeon:
Splits the current login page off to a modal, to be easier to track the user's current position to redirect back to on login finish. Also, moves the account settings to its new home in the new Settings redesign.


iameli (changes requested):
* Seems like when you log out you get dumped on an empty "Account" page
* Logged in --> Logged out --> Went to stream and now the button on the bottom is a permanent loading spinner (I guess this is a next thing too)

iameli (approved):
Looks great!

PR #748, opened 2025-11-26
feat: helper functions to fall back to DID if handle not available

### LINKED ISSUE #697: chat: if the user's handle is invalid, show DID instead of `@handle.invalid`
This is possibly fixed in chat but in general we should make sure the UX works and is not confusing. Easy to test: log in as a `did:web` without a valid handle and make sure stuff is okay

espeon:


js/components/src/components/share/sharesheet.tsx
```diff
@@ -63,31 +64,22 @@ export function ShareSheet({ onShare }: ShareSheetProps = {}) {
   // Share to Bluesky
   const shareToBluesky = useCallback(() => {
     const streamUrl = getStreamUrl();
-    const text = profile
-      ? `Check out @${profile.handle} live on Streamplace! ${streamUrl}`
-      : `Check out this stream on Streamplace! ${streamUrl}`;
+    const text =
+      profile && profile.handle
+        ? `Check out @${profile.handle} live on Streamplace! ${streamUrl}`
+        : `Check out this stream on Streamplace! ${streamUrl}`;
```
iameli, line 70: This one should be formatHandleWithAt right?

iameli (approved):
LGTM, thought I saw one spot you missed? Opened https://github.com/streamplace/streamplace/issues/752 'cause I had trouble logging in to test that but that's a super low-priority issue, don't block merging this one

js/components/src/components/share/sharesheet.tsx
espeon: I didn't use a formatting function there because I thought having the DID there would be confusing

PR #750, opened 2025-11-28
feat: polyfill date to sync client time

### LINKED ISSUE #663: normalize time client-side
Browser time being wrong has caused all sorts of unrelated issues, more recently this one but there have been others:

Image

We should look into using the server to normalize client-side time; I believe there are a variety of libraries for monkeypatching `Date` for this kind of thing

espeon:
have had some off/on client issues that could have been solved by syncing date. this includes a few hooks and similar to sync date on client init and polyfill date to account for the offset between client and server.


js/components/src/time-sync/time-sync.ts
```diff
@@ -0,0 +1,109 @@
...
+  return OriginalDate.now();
+}
+
+export function initializeTimeSync(): void {
+  if (Platform.OS !== "web") {
+    return;
+  }
+
+  // store original Date
+  OriginalDate = Date;
```
iameli, line 79: This makes me nervous because if initializeTimeSync gets called more than once we're gonna double-wrap it. Maybe put in a one-time static flag to check it's not called more than once?

js/components/src/time-sync/time-sync.ts
```diff
@@ -0,0 +1,109 @@
...
+      hasDrift: false,
+      driftMs: drift,
+      driftSeconds: Math.round(drift / 1000),
+    };
+  }
+}
+
+export function syncTimeWithServer(serverTime: string): void {
+  const serverDate = new Date(serverTime);
+  const clientDate = new Date();
```
iameli: This, right?

```suggestion
  const serverDate = new OriginalDate(serverTime);
  const clientDate = new OriginalDate();
```

pkg/spxrpc/place_stream_server.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"context"
+	"time"
+
+	placestreamtypes "stream.place/streamplace/pkg/streamplace"
+)
+
+func (s *Server) handlePlaceStreamServerGetServerTime(ctx context.Context) (*placestreamtypes.ServerGetServerTime_Output, error) {
+	serverTime := time.Now().UTC().Format(time.RFC3339)
```
iameli: I'm not 1000000000% sure of the distinction but I've been using `util.ISO8601` from "github.com/bluesky-social/indigo/util" lately just to match the conventions in indigo

js/components/src/time-sync/useTimeSync.tsx
```diff
@@ -0,0 +1,56 @@
...
+      if (Platform.OS !== "web") {
+        return;
+      }
+      try {
+        const agent = new StreamplaceAgent(url);
+        const response = await agent.place.stream.server.getServerTime();
+        const serverTime = response.data.serverTime;
+
+        // always sync with server time
+        syncTimeWithServer(serverTime);
```
iameli: So technically we can be ever-so-slightly more accurate here by assuming that the server generated its time halfway through the round-trip and compensating for that. It matters in the case where you're on a plane or something with working internet but 5000ms latency
```suggestion
        const agent = new StreamplaceAgent(url);
        const start = Date.getTime();
        const response = await agent.place.stream.server.getServerTime();
        const roundTripLatency = Date.getTime() - start;
        const serverTime = response.data.serverTime;

        // always sync with server time
        syncTimeWithServer(serverTime, roundTripLatency / 2);
```

iameli (changes requested):
Couple things to check and the round-trip-time thing

pkg/spxrpc/place_stream_server.go
espeon: totally fair, will change to match haha

PR #761, opened 2025-12-03
feat: typeahead in login

### LINKED ISSUE #754: login page: figure out some kind of autocomplete
People love it! Mixed feelings about just hitting the Bluesky appview for this, perhaps there's something better there. At a minimum we could resolve the handle once complete and show you your avatar so you're reasonably confident you're logging into the right thing.

espeon:
Should close #754. I tried to do it in a way that would make it harder for the average user to find specific handles if they didn't already know them beforehand. So, I ended up with more of an actual "autocomplete" system rather than a dropdown like most other apps would.
image


iameli (approved):
LGTM, maybe someday we index all the internet handles but w/e

PR #766, opened 2025-12-07
fix: infinite loop-ize ozone connecting

espeon:
still todo: set up a Grafana alarm for `labeler firehose connection has been failing for over a minute, Ozone may need a restart`

### DIFF
--- a/pkg/atproto/labeler_firehose.go
+++ b/pkg/atproto/labeler_firehose.go
@@ -23,7 +23,8 @@ import (
 
 func (atsync *ATProtoSynchronizer) StartLabelerFirehose(ctx context.Context, did string) error {
 	retryCount := 0
-	retryWindow := time.Now()
+	backoffDuration := time.Second
+	timeSinceStarted := time.Now()
 
 	for {
 		if ctx.Err() != nil {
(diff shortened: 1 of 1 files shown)


pkg/atproto/labeler_firehose.go
```diff
@@ -23,30 +23,42 @@ import (
...
+				"backoffDuration", backoffDuration,
+				"labelerDID", did)
+
+			// Wait with exponential backoff before retrying
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(backoffDuration):
+				// Exponential backoff with max of 5 minutes
+				backoffDuration = backoffDuration * 2
```
iameli, line 48: We've got a library I like for these already, see e.g. https://github.com/streamplace/streamplace/blob/c8583e1749c9eadaec5d5b961fbfef7ca7ce9cda/pkg/statedb/locks.go#L54-L71

5 minutes seems excessive; I can't imagine we're going to contribute to DDOSing anybody if it's a 30-second max (lots of places in the code already where 30 seconds is treated as an upper limit, such as "no segments found for 30 seconds, disconnecting ingest"

iameli (changes requested):
lil changes

espeon:
fixed in another PR :)

PR #768, opened 2025-12-08
fix: don't crash when labeler firehose is not available

### LINKED ISSUE #765: Don't crash when labeler firehose is unavailable
Set off alarms? Sure. Crashloop the node entirely? Nah.

ElshadHu:
### Description

I fixed an issue where labeler firehose connection failures could crash. I also added a Prometheus metric so I can observe labeler connections and implemented a graceful retry loop with backoff instead of crashing.

### Changes

- I added a `streamplace_labeler_firehoses_connected ` gauge metric in `spmetrics.go` to track labeler connections.
-  I used this metric to represent the connection state per labeler DID, where 1 means connected and 0 means disconnected.
-  I also changed the log level at the retry threshold from `Error` to `Warn`

### Testing

- I tested the change locally using Bluesky’s official moderation labeler
- I confirmed that the metric reports 1 while the labeler firehose is connected.
- Also I did the retry path and saw that the code backs off for 5 seconds after repeated failures instead of crashing the node.

Closes #765 

If there is any issue, please, let me know.


pkg/atproto/labeler_firehose.go
```diff
@@ -107,7 +110,8 @@ func (atsync *ATProtoSynchronizer) StartLabelerFirehoseRetry(ctx context.Context
 	if err != nil {
 		return fmt.Errorf("subscribing to firehose failed (dialing): %w", err)
 	}
-
+	spmetrics.LabelerFirehosesConnected.WithLabelValues(did).Set(1)
+	defer spmetrics.LabelerFirehosesConnected.WithLabelValues(did).Set(0)
```
iameli: We can connect to more than one labeler firehose, so this should be `Inc` and `Dec` instead of `Set`

iameli (changes requested):
Small change, looks okay

pkg/atproto/labeler_firehose.go
ElshadHu: Thanks! I changed to `Inc` and `Dec`

iameli (approved):

iameli:
thanks!

PR #770, opened 2025-12-09
feat: websocket like abstraction using iroh

rklaehn:
This replaces the explicit stream related logic with just a thin wrapper around an iroh endpoint/connection/stream pair that provides an interface similar to a websocket. All encoding and other logic would have to happen on the golang side.

Look at iroh_test/main.go for how this is used.

Questions:

- should we flatten connection and streams into a single thing, or expose connection?
- provide wrappers to go interfaces `Reader` and `Writer`. Especially Reader requires a tiny bit of boilerplate.

Note: we can't implement net.Conn because the close behaviour for QUIC connections is too different.

I think having send and recv side combined is the right choice, given that there are also some operations that affect the connection as a whole. The discussion would be more relevant if we would separate the concept of connection and stream in the binding.


rust/iroh-streamplace/src/c2pa.rs
```diff
@@ -1,286 +0,0 @@
-use std::{io::Cursor, sync::Arc};
```
iameli, line 1: We need this file, it does all of the c2pa signing! Admittedly it's incorrectly called "iroh-streamplace" when it's not Iroh but I don't want to go through the work of generating bindings for multiple Rust packages just yet

iameli (commented):
This looks conceptually reasonable but I won't have time to properly wire it up and test in the immediate future. If you or anyone else wants to take on the work, it'd be a matter of dumping the old `iroh_replicator` interface and using this to implement the Replicator https://github.com/streamplace/streamplace/blob/397d91a9f67296ce1c08fbcc0caee90273319158/pkg/replication/replicator.go#L11-L18

PR #774, opened 2025-12-11
feat: teleporting

espeon:
Introduces teleporting.

This introduces a basic slash-commands framework that hopefully should get more fleshed out in the future. Also introduces a chatbox-specific notification/toast component, but this time it's local to the chat. Perhaps both could be abstracted further to create a toast *primitive* though that both use.


iameli (changes requested):
LOVE the warning stripe effect! Just a little polish needed:

* [x] Looks like I'm always seeing "0 viewers" when I show up on the far end
* [x] 30 seconds is a bit long IMO. Let's maybe do 10; we could also have an optional second argument if someone wants to do `/teleport iame.li 60`
* [x] The destination URL is e.g. `/did%3Aplc%3Adkh4rwafdcda4ko7lewe43ml`; probably we should use a handle assuming one is available?
* [ ] The "cancel" button shows up to non-streamers and makes the toast go away and reappear (also it doesn't work)
* [x] You can teleport to yourself which is weird. Also it makes an infinite loop, which I guess is a very Portal thing to have happen to be fair. (Probably just self-teleports can be rejected on the frontend and indexer?)
* [ ] Stretch goal, entirely optional: Another warning-stripe effect for the actual transition? I know it'd be really hard to time correctly but a full-screen warning stripe appearing and then disappearing and you're on the other stream would go hard
* [ ] Doesn't seem to be working on mobile (tested on Android)

espeon:
fwiw - cancel button is for if the user doesn't want to teleport for whatever reason? i'm not 100% sure what the actual cancellation flow for the streamer looks like though

PR #778, opened 2025-12-12
fix: login on mobile failing metadata verification

espeon:
There was some sort of import conflict happening so I extracted the resolver so there isn't any. We can now log in successfully on mobile lol.

### DIFF
--- a/js/app/features/bluesky/oauthResolver.tsx
+++ b/js/app/features/bluesky/oauthResolver.tsx
@@ -1,0 +1,40 @@
+import type { OAuthAuthorizationServerMetadata } from "@atproto/oauth-client";
+import {
+  OAuthResolver,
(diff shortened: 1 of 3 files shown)


js/app/features/bluesky/oauthResolver.tsx
```diff
@@ -0,0 +1,40 @@
+import type { OAuthAuthorizationServerMetadata } from "@atproto/oauth-client";
+import {
+  OAuthResolver,
+  ResolveOAuthOptions,
+} from "@atproto/oauth-client/dist/oauth-resolver";
```
iameli, line 5: Guessing that the problem could be coming from is importing from `dist` here instead of the normal export? Regardless, happy to ship this if it works

iameli (approved):

PR #719, opened 2025-11-06
Delegated moderation lexicons

tripledoublev:
Implements delegated moderation lexicons as discussed in issue #718.

## New lexicons (`place.stream.moderation.*`):

| Type | Lexicon | Purpose |
|------|---------|---------|
| Record | `permission` | Grant moderation permissions to users |
| Procedure | `createBlock` / `deleteBlock` | Ban/unban users |
| Procedure | `createGate` / `deleteGate` | Hide/unhide messages |
| Procedure | `updateLivestream` | Update stream metadata |

## How it works

Moderators request actions → Streamplace validates permissions → actions execute using streamer's OAuth session →records stored in streamer's repo

---

Fixes #718 

This PR has other PRs merged into it, and will be merged last. It is the backend. See #796 for the frontend.

Issue #718, opened 2025-11-06
Delegated moderation actions


tripledoublev:
# Delegated Moderation Proposal

## Core Design

Moderators request actions → Streamplace validates permissions → actions execute using streamer's OAuth session → Records stored in streamer's repo.

**Benefits** (per [Eli's feedback](https://github.com/streamplace/streamplace/issues/718#issuecomment-3499684334)):
- Protects moderators (actions attributed to streamer)
- Actions persist (removing mod doesn't undo work)
- Reusable framework (chat + stream management)
- Simple architecture (no conflict resolution needed)

## Implementation

**Lexicons** (`place.stream.moderation.*`):

Record:
- `permission` - Grant moderation permissions to users

Procedures:
- `createBlock` / `deleteBlock` - Ban/unban users
- `createGate` / `deleteGate` - Hide/unhide messages
- `updateLivestream` - Update stream metadata

**Permission Scopes**:
- `ban` - User-level moderation
- `hide` - Message-level moderation
- `livestream.manage` - Stream metadata updates

<details>
<summary>Previous version (deprecated)</summary>
  
## Overview V2

**Goal**: enable streamers to delegate moderation powers to trusted users through a **delegated-action framework**.

**Core principle**: Moderators request actions → Streamplace validates permissions → Actions executed using **streamer's OAuth session** → Records stored in **streamer's repo**.

---

## Why This Approach

Based on [Eli's feedback](https://github.com/streamplace/streamplace/issues/718#issuecomment-3499684334):

1. **Protects moderators** - Actions attributed to streamer, not individual mods (with private audit log)
2. **Actions persist** - Removing a mod doesn't undo their past moderation work
3. **Reusable framework** - Same pattern for chat moderation AND stream management (titles, etc.)
4. **Simpler architecture** - No multi-mod conflict resolution, no complex authorization logic

---

### How It Works

```
1. Streamer creates delegation record → grants permissions to mod
   ↓
2. Mod requests action (ban user, hide message)
   → Calls Streamplace XRPC endpoint
   ↓
3. Streamplace validates permission
   → Uses streamer's OAuth session to create record
   ↓
4. Record created in STREAMER's repo
   → Private audit log tracks which mod requested it
   ↓
5. Enforcement: Check if record exists in streamer's repo
```

### Key design element

All records are in streamer's repo → Simple existence check → No conflicts possible

---

### Delegation (TBD naming)

Options: `place.stream.delegate` or `place.stream.delegation`

[...]

PR #787, opened 2025-12-16
feat(backend): Implement delegated moderation system

tripledoublev:
Implements backend for delegated moderation. 

## Components

- Database models (ModerationDelegation, ModerationAuditLog)
- Permission system with expiration enforcement
- Firehose integration for delegation records

## XRPC handlers
- `createBlock` / `deleteBlock` - Ban/unban users
- `createGate` / `deleteGate` - Hide/unhide messages
- `updateLivestream` - Update stream metadata

## Out of Scope (follow-up work)
- Audit log read endpoint for streamers
- Rate limiting

---

Depends on #719.
Part 2 of 4 for #718

### DIFF
--- a/pkg/atproto/sync.go
+++ b/pkg/atproto/sync.go
@@ -2,6 +2,7 @@ package atproto
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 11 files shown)


pkg/atproto/sync.go
```diff
@@ -423,6 +424,55 @@ func (atsync *ATProtoSynchronizer) handleCreateUpdate(ctx context.Context, userD
...
 
+	case *streamplace.ModerationPermission:
+		repo, err := atsync.SyncBlueskyRepoCached(ctx, userDID, atsync.Model)
+		if err != nil {
+			return fmt.Errorf("failed to sync bluesky repo: %w", err)
+		}
+		if repo == nil {
+			// someone we don't know about
+			return nil
+		}
```
iameli, line 435: Should be an impossible case if `SyncBlueskyRepoCached` just returned successfully
```suggestion
```

pkg/model/model.go
```diff
@@ -107,6 +107,17 @@ type Model interface {
 	CreateMetadataConfiguration(ctx context.Context, metadata *MetadataConfiguration) error
 	GetMetadataConfiguration(ctx context.Context, repoDID string) (*MetadataConfiguration, error)
 	DeleteMetadataConfiguration(ctx context.Context, repoDID string) error
+
+	CreateModerationDelegation(ctx context.Context, delegation *ModerationDelegation) error
+	DeleteModerationDelegation(ctx context.Context, rkey string) error
+	GetModerationDelegation(ctx context.Context, streamerDID, moderatorDID string) (*ModerationDelegation, error)
+	GetModerationDelegations(ctx context.Context, streamerDID, moderatorDID string) ([]*ModerationDelegation, error)
+	GetModeratorDelegations(ctx context.Context, moderatorDID string) ([]*ModerationDelegation, error)
+	GetStreamerModerators(ctx context.Context, streamerDID string) ([]*ModerationDelegation, error)
```
iameli, line 116: So most of Streamplace doesn't do this right now but I'm trying as much as possible to push the model layer to only ingest and return atproto structs from the `streamplace` package so we're not passing around `model` stuff elsewhere. That way the "model" indexing layer stays mostly self-contained. See the `BroadcastOrigin` stuff for a good example of how to do it

This might require to create a new non-record lexicon for "moderation delegation view" or something that wraps the record with an "author" but that'll be useful in other contexts too I think

pkg/model/model.go
```diff
@@ -107,6 +107,17 @@ type Model interface {
...
+
+	CreateAuditLog(ctx context.Context, log *ModerationAuditLog) error
+	GetAuditLogs(ctx context.Context, streamerDID string, limit int, before *time.Time) ([]*ModerationAuditLog, error)
+	GetModeratorAuditLogs(ctx context.Context, moderatorDID string, limit int, before *time.Time) ([]*ModerationAuditLog, error)
```
iameli, line 120: Still reading through but I don't think these can be in `model` -- that exists on every Streamplace node in parallel and should be mostly for records indexed from atproto. If we have secret multinode state that needs to go in `statefuldb` instead so it can be shared with multiple nodes through Postgres

iameli (changes requested):
I really like this! Proposed two changes:

1. I'm trying to move `model` away from returning its own objects in favor of returning atproto-compatible objects from the `streamplace` package whenever possible; this reinforces its role as an atproto indexer (with a couple narrow exceptions)
2. The audit log needs to move from `model` to `statefuldb` for multi-node reasons (but this should be relatively easy, the data model is otherwise fine)

3. EDIT: One more thing now that I've seen the whole system: I know "Edit livestream title" is how it looks to a user but we've actually moved away from ever editing `place.stream.livestream` records; we want the old records to still be there so they can be used as like "chapter markers" to annotate the recording of the stream. So let's have that feature create a new record instead, and for the initial version let's have it set `notificationSettings.pushNotification = false` (or whatever that field is called) so that mods don't start out with the permission to send new push notifications.

PR #790, opened 2025-12-16
feat: replace i18next-parser with i18next-cli

espeon:
i18next-parser was discontinued in september and they recommend switching


iameli (approved):
cool

PR #795, opened 2025-12-18
Sign segments with separate publisher key

### LINKED ISSUE #705: scope stream keys to the node that published them
Basically: currently `place.stream.key` records are global and unscoped. This means I could change my Streamplace node from `stream.place` to `streamplace.example.com` and stream in with the same key. This might seem like good UX, but comes with the huge downside that every Streamplace node that you ingest into could impersonate you to any other Streamplace node, which is really bad. If an attacker compromises `crappy-insecure-streamplace-node.com`, that shouldn't allow them to turn around and create a valid livestream for all of its users on `stream.place`.

The _reccomended_ soution for people that really care about this stuff is to run Streamplace Desktop locally adjacent to OBS and stream into that, using syndication to get the stream out onto the internet. That way the stream key stays fully custodial and never leaves your computer (and we can even use hardware signing with our PKCS#11 support, which I'm excited about.) But:

A. We're not ready with Streamplace Desktop and 
B. It's just not realistic for all use cases; doing local signing and broadcast from a phone with a crappy cell connection is a terrible idea.

So, the proposed solution: break apart the `c2pa.created` and `c2pa.published` actions. The creator signs a manifest using their stream key (either custodially or by including the private key in the OBS stream key field). Then, the broadcaster composes that segment and signs a `c2pa.published` action, using the first segment as an ingredient. **Importantly, both of these keys must be present in any Streamplace segment ingested by nodes. If you don't have both, you're looking at an unpublished segment; unpublished segments are implied to be personal-private and must not be distributed.**

So, some steps:

1. Create a new `place.stream.broadcast.publisherKey` record. The schema will look very similar to `place.stream.key`.
2. Add a `publisher` field to `place.stream.key` that is a reference to the DID. When we release this, all `place.stream.key`s without a publisher field will be deprecated.
3. While we're at it, let's pop an optional `expiresAt` field on both of those.
4. Teach Streamplace nodes to generate a publisher key (secp256k1) on boot and store in it statefuldb/publish it to the broadcaster's repository.
5. In the ingest pipeline, rewrite the signing process to do the double-signing scheme mentioned above.
6. In the validation pipeline, reject any segments that aren't double-signed. Validate that the `c2pa.published` action matches a valid, un-expired key, and that the `publisher` field of `place.stream.key` matches the identity that published the segment.

We have two more problems to solve here, but I'll write them up as separate tickets:

1. The `publisher` field of a `place.stream.key` record could be modified to point to somebody else, breaking all of the publisher-pinning semantics. To fix this we need to somehow add the publisher to the OBS stream key. So we'll figure that out.
2. The double-signing semantics make sense for livestreaming but archival is kinda weird. Currently, if a streamer deletes their `place.stream.key` record, we'll stop validating archived segments as "theirs". That's a little weird but probably okay. But a publisher deleting their `place.stream.broadcast.publisherKey` record shouldn't invalidate a users' segments? So we might need to figure out different publishing semantics around archival.

makew0rld:
Fixes #705 

**The PR builds on top of #716, so that PR must be merged before this diff can work.**

Still having some test issues, but we can resolve that later and in #705.


makew0rld:
Replaced by #807

PR #797, opened 2025-12-19
devex: use mise and confirm on CI build that mise works

espeon:


mise.toml
```diff
@@ -0,0 +1,7 @@
+[tools]
+go = "latest"
```
iameli, line 2: Hmm this isn't true exactly but I think it's fine, the latest version of Go should downgrade automatically to whatever we have in go.mod

iameli (approved):
Sweet

PR #798, opened 2025-12-20
feat: blur item if it's sensitive/objectionable

espeon:
Adds sensitive content blurring and warnings for streams labelled with certain content tags, via blurring thumbnails in the main live view, and a "sensitive content" overlay that obscures video and mutes audio until user acknowledgement.

Note that sensitive content overlay only darkens the background. Blurring would requite a new native dependency (expo-blur). We should also provide a method to show/blur/hide posts based on certain labels and other characteristics as well.


pkg/model/livestream.go
```diff
@@ -132,3 +135,22 @@ func (m *DBModel) GetLatestLivestreams(limit int, before *time.Time) ([]Livestre
 
 	return recentLivestreams, nil
 }
+
+func (m *DBModel) GetLivestreamView(ctx context.Context, ls *Livestream) (*streamplace.Livestream_LivestreamView, error) {
```
iameli, line 133: [Hmm same comment as I made here](https://github.com/streamplace/streamplace/pull/787/changes#r2636283706); it'd be cool to have the `model` functions just return a `*streamplace.Livestream_LivestreamView` in all cases. You don't gotta fully refactor that in this PR but maybe the new function should just take an ID and return a `*streamplace.Livestream_LivestreamView` without needing to perform an intermediate step elsewhere?

iameli (changes requested):
Broadly cool with this, just interested in simplifying `model` so we don't have both `model.Thing` and `streamplace.ThingView` structs running around everywhere.

An alternative to expo-blur would be doing a server-side blur on the thumbnail on the way out; potentially something that could be done in pure Go and cached so it'd be pretty lightweight

pkg/model/livestream.go
espeon: oh good point 😭 

Will see what I can do about doing that

espeon:
Probably not worth bothering atm until we get better moderation infrastructure

PR #799, opened 2025-12-20
fix: misc. navigation bar tweaks

espeon:
1. Logo in the top left redirects home, and we change the "Streamplace" on the navbar to "Home".
2. Adds a reset stack button to the 'settings' menu like we do already with 'home'
3. Adds a back button to non-base navigators (e.g. settings, home stacks)


iameli (approved):
sounds reasonable

PR #801, opened 2025-12-20
feat: s3 UI and saving to statedb

### LINKED ISSUE #576: s3: implement BYOS3 UX and statefuldb data model
Allow users to bring their own S3 credentials; when segmenting we'll upload everything there for archival

Could be in this format: `s3+https://ACCESS_KEY_ID:SECRET_ACCESS_KEY@s3.example.com/bucket`

espeon:
Adds basic s3 UI and data model per #576


pkg/statedb/storage.go
```diff
@@ -0,0 +1,95 @@
...
+	streamplace "stream.place/streamplace/pkg/streamplace"
+)
+
+// Storage represents S3 storage configuration for a user
+type Storage struct {
+	ID                         string    `gorm:"column:id;primarykey"`
+	IsActive                   bool      `gorm:"column:is_active;default:true"`
+	UserDID                    string    `gorm:"column:user_did;not null;unique"`
+	URL                        string    `gorm:"column:url;not null"`
+	RequestedSecondsPerSegment int       `gorm:"column:requested_seconds_per_segment;default:6"`
```
iameli: This is interesting but maybe leave out for now; we'll want to start with large archival files

espeon: Good point, was thinking about splitting up into slightly larger (10s instead of 1s) segments when writing that.

iameli:
@espeon Obviously this doesn't do much until I implement archival (which probably means fixing deterministic muxing) but rather than let the PR sit around getting stale I'm kind of tempted to ship it and hide it from the UI? Just so you don't gotta keep rebasing

espeon:
this is doable :0

iameli (approved):
LGTM, let's get this merged but hidden in the UI so it's ready when the time comes

PR #802, opened 2025-12-27
feat: bump deps and add liquid glass (lol)

espeon:
Not 100% sure about this one - had to update react-navigation to v8-beta, but it works like a charm for now 🤔. If needed, i can definitely remove and go back to an earlier react navigation version, but we wouldn’t be able to keep the liquid glass nav bar.

There was also a refactor of the main router, so I’ll have to check and make sure all the background features work as well as they did beforehand.

IMG_1857


espeon:
closing in favour of #940

PR #820, opened 2026-01-09
fix: block javascript URIs (they are most likely malicious anyways)

### LINKED ISSUE #819: Refuse to index `place.stream.chat.message`s with `javascript:` facets
[So this is an attack that hit some people today:](https://bsky.app/profile/skywatch.blue/post/3mbw62gvobk2r) pretty simple one, just a clickable link has `<a href="javascript:evilCodeHere()">` that's not getting sanitized properly. We appear to have not been susceptible but any chat posts that have that are pretty much malicious by definition and we should do defense-in-depth by refusing to index them.

espeon:
should resolve #819

### DIFF
--- a/pkg/atproto/sync.go
+++ b/pkg/atproto/sync.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


iameli (approved):
Sweet

PR #821, opened 2026-01-09
fix: astro uses global favicon

### LINKED ISSUE #813: docs: fix favicon to not be astro default

espeon:
Should close #813


iameli (changes requested):
This'll be the generic pink cube which is not what we're using in prod right now. Is it possible to tell Astro to not use any favicon?

espeon:
the web service should serve up the node's favicon as /favicon.ico afaik, pdsls.dev as an example uses it fine
image

iameli:
Oh sweet, sure

PR #824, opened 2026-01-10
fix: add overflow hidden to stream monitor to fix square corners

### LINKED ISSUE #818: Square corners in live dashboards
some square corners instead of round ones in the live dashboard:

Image

ElshadHu:
## Fix
fix

I considered adding border-radius to each child component individually, but that would require modifying multiple children . Instead, I checked out the pattern already established in the codebase by checking `cards.tsx` and `avatar.tsx` where overflow: "hidden" is used on the parent . I think using overflow: "hidden"  keeps the fix simple . 
Fixes #818


iameli:
Cool, thanks!

PR #826, opened 2026-01-10
feat: move log in/sign up to top right

espeon:
Suggestion from @uncenter, but I personally like it as well. Most people expect the buttons to be near the avatar wherever it may be, so it's probably best ux-wise to add links around to log in/sign up.

image

### DIFF
--- a/js/app/components/login/login-form.tsx
+++ b/js/app/components/login/login-form.tsx
@@ -74,7 +74,8 @@ export default function LoginForm({ onSuccess }: LoginFormProps) {
   };
 
   const onSignup = () => {
-    loginAction("https://bsky.social", openLoginLink);
+    // TODO: remove requirement for oauth-protected-resource in oatproxy
+    loginAction("https://blewit.us-west.host.bsky.network", openLoginLink);
   };
 
   const isMobile = Platform.OS === "ios" || Platform.OS === "android";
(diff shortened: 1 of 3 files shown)


js/app/components/login/login-form.tsx
```diff
@@ -74,7 +74,8 @@ export default function LoginForm({ onSuccess }: LoginFormProps) {
   };
 
   const onSignup = () => {
-    loginAction("https://bsky.social", openLoginLink);
+    // TODO: remove requirement for oauth-protected-resource in oatproxy
+    loginAction("https://blewit.us-west.host.bsky.network", openLoginLink);
```
iameli: I don't get it

js/app/src/router.tsx
```diff
@@ -293,36 +292,111 @@ const NavigationButton = ({ canGoBack }: { canGoBack?: boolean }) => {
...
+
+    const handleResize = () => {
+      setWindowWidth(window.innerWidth);
+    };
+
+    window.addEventListener("resize", handleResize);
+    return () => window.removeEventListener("resize", handleResize);
+  }, []);
+
+  const isCompact = windowWidth <= 800;
```
iameli: Does the React Native `Dimensons` stuff not help here? Native width can change too via like iPad split-screen

js/app/components/login/login-form.tsx
espeon: tbh probably better to address the actual issue vs monkeypatching it like this, will take it out for now

js/app/src/router.tsx
espeon: Good point, useWindowDimensions exists

iameli (approved):

PR #840, opened 2026-01-14
feat: wrangler.toml for static docs deployment

espeon:


js/desktop/package.json
```diff
@@ -11,7 +11,7 @@
     "make": "electron-forge make",
     "publish": "electron-forge publish",
     "nolint": "eslint --ext .ts,.tsx .",
-    "prepare": "CGO_ENABLED=0 go run ../../pkg/config/git/git.go --js > src/version.ts"
+    "prepare": "if [ -z \"$CF_PAGES\" ]; then CGO_ENABLED=0 go run ../../pkg/config/git/git.go --js > src/version.ts; else echo 'export const version = \"unknown\"; export const buildTime = 0; export const uuid = \"00000000-0000-0000-0000-000000000000\";' > src/version.ts; fi"
```
espeon: this is only for desktop but i'd really like for us to not need to do this eventually? glad that nothing else js-wise hits it though.

iameli: We can have this same logic but just do it inside `git.go`, bit cleaner

pkg/config/git/git.go
```diff
@@ -84,7 +84,21 @@ func makeGit() error {
 	flag.Parse()
 	r, err := git.PlainOpenWithOptions(".", &git.PlainOpenOptions{DetectDotGit: true})
 	if err != nil {
-		return err
+		// fallback for CI environments without full git history
```
iameli: Should just check for the env var here

iameli (approved):

PR #841, opened 2026-01-14
Fix char limit chat

### LINKED ISSUE #806: Chat : Character limit
In the process of chatting during a live stream the chat messages appear to not have a character limit
i tested with a 300 character, then 1000 character, then 2000 character lorem ipsum

Thoughts:

- to avoid a wall of text filling chat, [Copy Pasta](https://www.twitchquotes.com/copypastas/labels/classic), a character limit needs to be established
- for interoperability with Blusky / Skylight / Roomy.Chat / Wisp.place and future integrations, there should be a character limit so the other sites are not overwhelmed with a text block
- moderation / labeling will also need to be a consideration with walls fo text, and ascii art

ElshadHu:
## Backend
I added message truncation to  `ToStreamplaceMessageView()`
- if a chat message exceeds 300 characters, it gets trimmed to 300. This acts like if any messages that bypass frontend validation.

## Frontend 

I added a check before sending  in `chat-box.tsx`
- if the message is over 300 chars, show an error toast and block submission . I added a red border on the text input when you're over the limit, so users know immediately they've hit it before even they are trying to send it.

## Screenshots
image
image

Closes #806


js/components/src/components/chat/chat-box.tsx
```diff
@@ -234,6 +236,17 @@ export function ChatBox({
 
   const submit = () => {
     if (!message.trim()) return;
+    if ([...message].length > 300) {
```
espeon: This is probably fine, but I think the correct way here is to use `Intl.segmenter()`

```suggestion
    if ([...new Intl.Segmenter().segment(message)].length > 300) {
```

js/components/src/components/chat/chat-box.tsx
```diff
@@ -62,6 +63,7 @@ export function ChatBox({
     new Map(),
   );
   const [filteredEmojis, setFilteredEmojis] = useState<any[]>([]);
+  const isOverLimit = [...message].length > 300;
```
espeon: same as below 

```suggestion
  const isOverLimit = [...new Intl.Segmenter().segment(message)].length > 300;
```

espeon (changes requested):
Good so far! Probably fine to merge it right now but had one nit 🥲

ElshadHu:
Thanks @espeon 🙌, I will check it out

espeon (approved):
Looks good, tested on web + iOS and it seems to work great.

Made a commit to go back to the spread string method as Intl.Segmenter didn't work on native platforms :(

ElshadHu:
> Looks good, tested on web + iOS and it seems to work great.
> 
> Made a commit to go back to the spread string method as Intl.Segmenter didn't work on native platforms :(

I didn't know it wasn't supported on native platforms. Thanks for review and testing 🫡

pkg/model/chat_message.go
```diff
@@ -41,6 +41,14 @@ func (m *ChatMessage) ToStreamplaceMessageView() (*streamplace.ChatDefs_MessageV
 	if err != nil {
 		return nil, fmt.Errorf("error decoding feed post: %w", err)
 	}
+	// Truncate message text if it is a ChatMessage
+	if msg, ok := rec.(*streamplace.ChatMessage); ok {
+		runes := []rune(msg.Text)
+		if len(runes) > 300 {
+			msg.Text = string(runes[:300])
+		}
+	}
```
iameli, line 57: I had to go read up on this, but it turns out that this will iterate over Unicode code points, while the thing we actually want to enforce is that it's >= 300 Grapheme clusters. Bluesky's Go code seems to use this library for it https://github.com/rivo/uniseg

js/components/src/components/chat/chat-box.tsx
iameli: Turns out this doesn't work in React Native but the `graphemer` npm package will work `import Graphemer from 'graphemer'`

pkg/model/chat_message.go
ElshadHu: @iameli  thanks, I'll check out uniseg

ElshadHu:
I tested with various emojis such as family emoji (👨‍👩‍👧‍👦) , country flags and it now limits based on grapheme clusters instead of Unicode code. I couldn't find a built-in truncation method in `uniseg` so I wrote the iteration. Also I noticed `graphemer` is archived but it seems stable, hopefully that's not an issue for dependencies.

ElshadHu:
Without frontend validation, the old backend truncation would cut in the middle of family emojis, breaking them into separate person emojis
image

iameli (approved):

PR #842, opened 2026-01-15
feat: analytics service

espeon:
Establish core architecture for the analytics service, including a grpc interface for communication between this service and the AppView.

Highlights:
- Clickhouse main analytics storage
- reDB based on-disk WAL (flushes to clickhouse every 1k events or 5 seconds)
- AppView access via grpc (we can add an Axum router if needed for un/authenticated read?)
- prometheus (ofc)

todo:
- [ ] opt-out (mention that opting out will send a less # of anonymised requests used only for QOS)
- [ ] gated xrpc query endpoints for accessing statistics
- [ ] flesh out deletion requests more (automatic? manual?)
- [ ] backfilll tooling for atproto-record based data (nexus/tap)
- [ ] MVs?


js/components/src/components/ui/menu.tsx
```diff
@@ -285,63 +279,63 @@ export const MenuInfo = forwardRef<View, MenuInfoProps>(
   },
 );
 
-export interface MenuDraggableGroupProps {
```
espeon, line 288: weeee should just delete these

espeon:
will close this for now, as it seems like it won't be merged anytime soon

PR #847, opened 2026-01-16
fix: misc moderation changes

### LINKED ISSUE #839: have loader underneath the UI
the loader blocks us from accessing settings, so we'll have to enable clickthrough or have the loader underneath the UI so we can e.g. disable low latency mode.

espeon:
1. should fix #839 
2. adds a kebab icon with report sections on desktop
3. allows chat menu when not logged in
4. finally fix report box buttons


iameli (changes requested):
I think having both "report livestream" and "report user" is kinda confusing but I'd be cool with just sending account-level stuff, unless we're truly truly just talking about reporting the contents of the livestream record e.g. the stream title

PR #849, opened 2026-01-16
Mute player in live stream on dashboard

ThisIsMissEm:
I noticed today that whilst setting up my stream I had a horrible feedback loop, this was caused by the player on the live dashboard being unmuted. This change *should*, if I'm right, mute that player by default.


iameli:
@ThisIsMissEm Looks like we're not respecting that property right now :( If we _were_, we'd also want the ability to unmute from the dashboard so you can verify that your audio works. @espeon will poke into why we're not respecting it

ThisIsMissEm:
Oooh, alrighty, shall I close this or keep it open?

espeon:
will close this and make my own pr 🫡

PR #853, opened 2026-01-17
Background workers for status and origin updates

jazware:
Don't block on status updates, don't spawn new goroutines for them every segment, coalesce notifications to update status (so we don't queue them up).


jazware:
Not totally sure about your shutdown logic for a stream, if the context shutdown mechanism works that works but I added a channel we can close to notify the background routines.

iameli (approved):
LGTM!

PR #865, opened 2026-01-19
feat: restore last metadata

espeon:
Fill in the last title if available and provide a button to fetch the last stream image


js/app/components/live-dashboard/livestream-panel.tsx
```diff
@@ -338,6 +365,62 @@ function LivestreamPanel({ scrollable = true }: { scrollable?: boolean }) {
...
+      } else if (did.startsWith("did:plc:")) {
+        // For did:plc, use plc.directory
+        const didResponse = await fetch(`https://plc.directory/${did}`);
+        if (!didResponse.ok) {
+          throw new Error("Failed to resolve DID document");
+        }
+        didDoc = await didResponse.json();
+      } else {
+        throw new Error(`Unsupported DID method: ${did}`);
+      }
```
iameli: Surely we have reusuable code for handle resolution somewhere already?

espeon: I thought we did too but I can check again? i remember seeing one somewhere but that only supported PLC

iameli: If not this should be put somewhere where it can be reused

iameli: One other place we do it but ideally these would get unified: https://github.com/streamplace/streamplace/blob/833a9ff7142f48c68f6415b0b9b3228392227892/js/app/store/slices/contentMetadataSlice.ts

espeon: On it!

iameli (approved):

PR #866, opened 2026-01-19
docs: remove accidentally-commited generated docs

iameli:
Also update `make lexicons` to regenerate the Markdown directory so this doesn't happen again

We don't want to just put those files in `.gitignore` because it's helpful to have them there for a fast docs deploy


espeon (approved):
waow lgtm i like this!

Issue #784, opened 2025-12-15
Feature Request / Composable Moderation : honor blocks / mutes / viewable status from Bluesky


PSingletary:
The moderation of a streamchat needs to serve multiple purposes:

- Broadcaster moderating their content, which chat messages are a component of streams
- Blocks and mutes need to be honored in chat of any stream place stream
- Messages being visible to logged out users needs to be honored in chat of any stream place stream

This request may need to be broken down into different components to develop

makew0rld:
Transforming this into more detailed feature requests: 
- [x] Streamers moderating chat msgs - #718 
- [ ] Chatters having their bsky blocks and mutes honoured from their view of chat and the homepage - #873 
- [ ] ~~Chat messages not visible publicly if chatter has `!no-unauthenticated` label on the bsky profile~~

makew0rld:
Questions:
- Should a viewer not be able to see a streamer they've blocked, even when navigating to their page directly?
- Should a viewer not be able to _view_ a stream if the streamer has blocked them? (even though chats would be hidden)

makew0rld:
Logged out users questions:
- Does it make sense for a user's settings on logged-out users on Bluesky to be mirrored in Streamplace? Is this violating the principle of least surprise?
- Should there be a streamplace-specific override?
- Does having logged-out non-visibility not make sense at all?
  - Streamers often have all the chat msgs in the video stream
  - There isn't a social expectation of chat msgs being private or hidden
  - The conversation is broken up if you're not logged-in

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/10f76359-7e65-41a7-a7bc-8b9ba1d2f27a" />

iameli:
Yeah, I don't think I'd want to proactively jump on that one — we've never had this guarantee and nobody's yet asked us for it. For Bluesky it was a compromise to ease the transition from the invite-only universe, we don't really have that

PR #873, opened 2026-01-21
feat: filter blocked users from home feed and chat

makew0rld:
Add client-side filtering based on the viewer's Bluesky block list:
- Fetch block records using com.atproto.repo.listRecords on login
- Subscribe to real-time block/unblock events via Jetstream
- Filter blocked users' streams from the home screen
- Filter blocked users' messages from chat
- Uses @skyware/jetstream for real-time updates when blocks are created or deleted.

Addresses bullet point 2 in #784

I opted not to address mutes, as I feel like mutes might be used more for bluesky-specific content and so filtering atproto content out there would be surprising or incorrect. Let me know what you think. Also mutes would require polling since they are private.

Notably, I failed to implement code that would use `app.bsky.graph.getBlocks`. I just couldn't get the oauth to work, even after adding a scope and proxied streamplace XRPC pathway for it. Let me know if you have any insight. But since all we need are DIDs, using listRecords is more efficient anyway.

### DIFF
--- a/js/components/src/livestream-provider/websocket.tsx
+++ b/js/components/src/livestream-provider/websocket.tsx
@@ -1,10 +1,12 @@
 import { useRef } from "react";
 import useWebSocket from "react-use-websocket";
 import { useHandleWebsocketMessages } from "../livestream-store";
-import { useUrl } from "../streamplace-store";
+import { useDID, useStreamplaceStore, useUrl } from "../streamplace-store";
 
 export function useLivestreamWebsocket(src: string) {
   const url = useUrl();
+  const did = useDID();
+  const oauthSession = useStreamplaceStore((state) => state.oauthSession);
   const handleWebSocketMessages = useHandleWebsocketMessages();
 
   let wsUrl = url.replace(/^http\:/, "ws:");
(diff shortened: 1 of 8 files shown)


iameli (changes requested):
I don't think we want to add another external dependency on Jetstream here or hit the Bluesky XRPC endpoint, [we're already indexing everyone's blocks in the local database here](https://github.com/streamplace/streamplace/blob/5ea7ad20a8cbf83a0f554ba1348972cc34d9b518/pkg/atproto/sync.go#L59-L84). So ideally this could be enforced on the AppView layer by just not sending messages over the websocket when there's a block happening etc. We do that already for the streamer, just need to honor it for other viewers as well. We can do that in the websocket, though we may have to handle it in both the [initial burst](https://github.com/streamplace/streamplace/blob/5ea7ad20a8cbf83a0f554ba1348972cc34d9b518/pkg/api/websocket.go#L235-L242) and the [following live message feed](https://github.com/streamplace/streamplace/blob/5ea7ad20a8cbf83a0f554ba1348972cc34d9b518/pkg/api/websocket.go#L283-L288).

I guess to do that you'll need to teach the websocket which user, if any, is currently logged in; let me know if there's trouble there and I'll help you out

makew0rld:
That's a good suggestion, thank you. I've made these changes so the only dependency is the node.

The only difference in UX with this new changeset: unblocks are not reflected in chat until reload. I think that's acceptable.

js/components/src/livestream-provider/websocket.tsx
```diff
@@ -15,7 +17,14 @@ export function useLivestreamWebsocket(src: string) {
   const hasReceivedMessage = useRef(false);
   const hasErrored = useRef(false);
 
-  const { readyState } = useWebSocket(`${wsUrl}/api/websocket/${src}`, {
+  // Don't connect until auth state is resolved (undefined = still loading)
+  const authResolved = oauthSession !== undefined;
```
iameli, line 21: we have a hook specifically for this use case, `import { useIsReady } from "store/hooks";`. Handles some corner cases

makew0rld: Having looked at this more, is keeping things as it is maybe more efficient? Since `useIsReady` requires the full profile is loaded. It's also difficult to use that function as it would cause a circular import.

PR #879, opened 2026-01-21
xrpc: implement com.atproto.sync.getRepo

iameli:


G34hWeatherSpotter:
i think you took down bsky 
image

Issue #823, opened 2026-01-09
Enhancement : SmokeSignal.events interoperability


PSingletary:
A smokesignal.events for streamplace streams should auto notify everyone that RSVP'd as yes 

- Smokesignal organizer connected to streamplace channel live detection
- Smokesignal RSVP yes/interested + opt in to notification
- Streamplace go live notice, the manual one from dashboard, should provide options to notify Smokesignal RSVP AND opt in to notification to mention all participants the stream has started

PSingletary:
https://bsky.app/profile/psingletary.com/post/3mbz4nsgyxk23

PSingletary:
https://bsky.app/profile/psingletary.com/post/3mbz4nsp6ok23

tripledoublev:
I've implemented Smokesignal interoperability in PR #882.

#### What's Implemented
A **SmokesignalEventForm** component in the livestream dashboard that uses Smokesignal's [Action Intent Links](https://discourse.smokesignal.events/t/action-intent-links/172) feature (released Jan 13, 2026). Streamers can pre-fill event details (title, description, date/time) and the stream URL is automatically included. Clicking `Create SmokeSignal event` opens smokesignal.events to create the event and auto-populates the form's field.

#### Limitations of This Approach
1. **Opens new browser window** - users leave Streamplace
2. **Login state dependency** - if not logged into smokesignal.events, URL params are lost
3. **No programmatic feedback** - this doesn't confirm the event creation or retrieve the event AT-URI

#### Proposed Future Phases

##### Phase 2: Direct Lexicon Record Creation
Instead of redirecting, Streamplace could create `community.lexicon.calendar.event` records directly to the user's PDS:

##### Pros:
-  Users stay in Streamplace
- We control the full flow
- Smokesignal indexes the `community.lexicon.calendar.*` record 

##### Cons:
- Users would either lose access to advanced smokesignal.events features
- To support more features, we would find ourselves replicating the smokesignal form interface

#### Phase 3: Consent-Based Notifications
To address the original issue's goal of notifying RSVPs when going live:

1. **New subscription record**: A `place.stream.notification.subscription` record where users opt-in to mentions:
   ```json
   {
     "$type": "place.stream.notification.subscription",
     "channel": "bluesky",
     "allowMentions": true,
     "createdAt": "..."
   }
   ```
[...]

PR #882, opened 2026-01-23
feat: SmokesignalEventForm - Action Intent Links integration

tripledoublev:
## Summary
- Adds SmokesignalEventForm component to the livestream dashboard
- Enables streamers to create Smokesignal events with pre-filled stream details
- Uses Smokesignal's Action Intent Links feature (similar to Bluesky's intent compose links)

## Changes
- New `smokesignal-event-form.tsx` with form validation
- Integration in `livestream-panel.tsx`

## Limitations & Future Work
This is a **Phase 1 implementation** using Action Intent Links. See #823 for discussion of future phases including:
- Direct `community.lexicon.calendar.event` record creation
- Notification consent records for opt-in mentions
- RSVP-based attendee notifications

## Testing

When testing, if the user isn't logged in, the login process will wipe the URL params used by the  Action Intent Links feature. This is the biggest reason for creating `community.lexicon.calendar.event` records directly from within Streamplace.

Relates to #823

### DIFF
--- a/js/app/components/live-dashboard/livestream-panel.tsx
+++ b/js/app/components/live-dashboard/livestream-panel.tsx
@@ -1,3 +1,4 @@
+import { useNavigation } from "@react-navigation/native";
 import {
   Admonition,
   Button,
@@ -19,7 +20,7 @@ import {
   useUrl,
   zero,
 } from "@streamplace/components";
-import { ArrowRight, ImagePlus, X } from "lucide-react-native";
+import { ArrowRight, ChevronRight, ImagePlus, X } from "lucide-react-native";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import {
   Image,
@@ -32,7 +33,6 @@ import {
 import { useUserProfile } from "store/hooks";
 import { useCaptureVideoFrame } from "../../hooks/useCaptureVideoFrame";
 import { useLiveUser } from "../../hooks/useLiveUser";
-import { SmokesignalEventForm } from "./smokesignal-event-form";
 
 const { flex, p, px, py, gap, layout, bg, borders, text, r, w, typography } =
   zero;
@@ -212,6 +212,7 @@ function LivestreamPanel({ scrollable = true }: { scrollable?: boolean }) {
   const createStreamRecord = useCreateStreamRecord();
   const updateStreamRecord = useUpdateStreamRecord();
   const url = useUrl();
+  const navigation = useNavigation();
 
   const [title, setTitle] = useState("");
   const [loading, setLoading] = useState(false);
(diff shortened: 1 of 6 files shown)


iameli (changes requested):
Great start! Got it working with exactly the friction you described, had to log into Smokesignal and then do the link again.

@tripledoublev feedback:

* Let's move this to a dedicated screen, I don't really think there's a lot of virtue in adding yet another thing to the live dashboard for this use case. You can totally link to it from the live dashboard though.
* The little "time zone" blurb turned out to be a lie - my test account's Smokesignal time zone was set to America/New_York instead of America/Los_Angeles. That might not be possible to solve through the intent link
* Possible we can use actual date/time pickers instead of typing the date and time manually? Surely React Native ships something for this, right?

Smokesignal feedback:

* Sooo the login flow seems to eat the intent link. But we have the users' handle, so is it possible we could include that in a query parameter and then bounce them directly to an OAuth consent screen? That seems like the smoothest way to get people logged in here.

ngerakines:
Can I get a couple of example URLs that you're building? I was able to use the create event intent link without issue when unauthenticated.

tripledoublev:
Thanks for the feedback @iameli.

- Added a SmokeSignalEvent screen at /smokesignal-event instead of embedding the form in the live dashboard.
- Replaced date/time inputs with the following pickers:
  - Web: HTML5 <input type="date"> and <input type="time">
  - Native: @react-native-community/datetimepicker

**OAuth / params preservation**: Smokesignal’s login flow drops the intent link. I tested it by including the handle in a query param, but params could not be preserved through OAuth. I'm probably just missing the proper approach on the Smokesignal side.

here's an example URL I'm building:                                     
                                                                                       
https://smokesignal.events/event?name=Test%20Livestream&description=Live%20on%20Streamplace%3A%20https%3A%2F%2Fstream.place%2Ftest&mode=virtual&starts_at=2026-02-03T18%3A00%3A00.000Z&ends_at=2026-02-03T20%3A00%3A00.000Z&link=https%3A%2F%2Fstream.place%2Ftest&link_name=Watch%20on%20Streamplace                                               
                                                                                       
When I open this while unauthenticated, I get redirected to `Sign in to Smoke Signal`. After login, I was sent to the homepage with params gone.

I found the OAuth destination column is `VARCHAR(255)`, so long URLs get truncated and the redirect fails. I’ve implemented a workaround:

- omit the link URL when the destination would exceed the limit and show a hint to user to add it manually in Smoke Signal after creating the event.

Question for @ngerakines, would it be possible to extend the destination column limit so longer intent URLs can be preserved through OAuth?

espeon:
@react-native-community/datetimepicker is a native dependency, so it'll require a new native build and can't be OTAed. This is fine (we should probably get a new native build out soon) but just something to keep in mind.

ngerakines:
> Question for @ngerakines, would it be possible to extend the destination column limit so longer intent URLs can be preserved through OAuth?

Yep, I can get that out today.

PR #887, opened 2026-01-27
feat: add a "stop stream" button

### LINKED ISSUE #884: mobile: add "STOP STREAMING" button
And lock people inside the streaming interface until they're done

espeon:
Stop stream button is in the streamer context menu (shows up if one is actively streaming from the player), and has a widget that pops out of the menu saying "stop stream here".


iameli:
Good 'nuff for now

PR #893, opened 2026-01-30
fix: remove stream viewers from header as it's just duplicate data

### LINKED ISSUE #836: hide user count by clicking
We do it in one place but not in another on the live dashboard, would make sense that both are saved and synced

espeon:
closes #536 by sidestepping around it. Only issue is there's nothing there anymore when live.


js/components/src/components/dashboard/information-widget.tsx
```diff
@@ -38,7 +39,7 @@ export default function InformationWidget({
   const [bitrateHistory, setBitrateHistory] = useState<number[]>(
     Array.from({ length: BITRATE_HISTORY_LENGTH }, () => 0),
   );
-  const [showViewers, setShowViewers] = useState(false);
+  const [showViewers, setShowViewers] = useAQState("showViewers", false);
```
iameli: I think
```suggestion
  const [showViewers, setShowViewers] = useAQState("showViewers", true);
```
I assert wanting to know is the default

PR #896, opened 2026-02-02
fix: support all ATProto lexicon datetime formats in aqtime

makew0rld:
I noticed errors in my node's log around timestamps. This change supports all the valid timestamps from [the spec](https://atproto.com/specs/lexicon#datetime) while rejecting invalid ones. `time.Parse` is relied on so this should be fast and correct.


iameli (approved):
LGTM but restarted some CI jobs, will merge when passing

PR #897, opened 2026-02-03
feat: add branding tags on web

### LINKED ISSUE #895: "My Streamplace Station" shows up for a second before custom branding
Both in the <title> and the corner of the screen. We could probably fix this with another HTML injection hack where we ship the necessary branding details in a <script> before loading the rest of the app?

espeon:
Adds a few headers on web to help with preloading branding, leading to less content flash.

image


iameli (changes requested):
Looking better, let's just do the `<title>` tag too and I think we'll be good to go `<title>Stream.place</title>`

PR #898, opened 2026-02-03
fix: offset 'scroll to bottom' when closed to not interfere with chat

### LINKED ISSUE #877: Bug: Scroll to bottom non-functional and blocking chat in Firefox
The scroll to bottom element on Firefox is blocking chat, preventing clicking links sometimes.

https://youtu.be/Gq5vOpTUw4w

espeon:
Offset it to the bottom a little. Previously, it blocked the last line of chat such that no links were clickable, etc.
image


iameli (approved):
sho

PR #902, opened 2026-02-05
fix: Restrict returnRoute (login redirection) handling to web

espeon:
I thought this already existed but I guess not

### DIFF
--- a/js/app/features/bluesky/blueskyProvider.tsx
+++ b/js/app/features/bluesky/blueskyProvider.tsx
@@ -24,7 +24,9 @@ export default function BlueskyProvider({
     loadOAuthClient();
 
     // load return route from storage on mount
-    if (Platform.OS !== "web") return;
+    if (Platform.OS !== "web") {
+      return;
+    }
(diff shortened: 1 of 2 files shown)


js/app/features/bluesky/blueskyProvider.tsx
```diff
@@ -23,6 +24,7 @@ export default function BlueskyProvider({
     loadOAuthClient();
 
     // load return route from storage on mount
+    if (Platform.OS !== "web") return;
```
iameli: Personal preference but I don't really like inline if here, I wonder if there's a prettier setting for it

```suggestion
    if (Platform.OS !== "web") {
      return;
    }
```

iameli (approved):
LGTM, formatting preference, feel free to merge tho

js/app/features/bluesky/blueskyProvider.tsx
espeon: ooh good to know, i can look into that

PR #907, opened 2026-02-10
feat: move to your favourite cli library

espeon:
Move to urfave/cli, primarily for maintenance reasons with CLI submodules. Should act 100% the same afaict.


iameli (changes requested):
Gotta migrate all the subcommands over too, I think `streamplace whip` and `streamplace whep` aren't working 'cause they use old flags

iameli:
But I think I fixed the Livepeer stuff

espeon:
Should be all the submodules that would need updating? Not sure if I'm missing anything

iameli (approved):

PR #909, opened 2026-02-11
feat: teleport dialog!

espeon:
When doing `/teleport` with not directing who to target, we now pull up a teleport dialog so one can choose where one wants to teleport.

image


espeon:
Unrelated but I really want to redo a lot of the stuff relating to dialogs. Especially with the new signup flow, I just tend to ditch most of the components I made because they just don't fit right.

iameli (approved):
siiiiick

maybe a button for it someday

PR #910, opened 2026-02-12
feat: mod, VIP badges

espeon:


espeon:
Was going to do custom badge creation + verification but i think for now that may be too much work atm

lexicons/place/stream/badge/display.json
```diff
@@ -0,0 +1,44 @@
...
+        "required": ["badges"],
+        "properties": {
+          "badges": {
+            "type": "array",
+            "description": "Up to 3 badge tokens to display with the message. First badge is server-controlled, remaining badges are user-settable.",
+            "maxLength": 3,
+            "items": {
+              "type": "ref",
+              "ref": "#badgeSelection"
+            }
```
iameli: Seems like this would need a streamer field?

iameli (changes requested):
* Let's lose issuance for now — I want it too but I wouldn't want to run out in front of emotes
* That should mean we don't need any new records for the moment IIUC
* I like the purple background but on the live badge the contrast ratio from the purple to the red is hurting my eyes a little
* These maybe need to be twice their resolution for retina display stuff yeah?

iameli (approved):
Image Cool with this assuming the jagged edges are a design choice, if not they should be smooooth

PR #911, opened 2026-02-12
feat: resolution display for source in source selection

espeon:
image


espeon:
Thinking we should round that last parameter unless we have developer mode on haha

iameli (approved):

PR #923, opened 2026-02-14
feat: add headphone play/pause button support

JordanWeatherby:
Allow pausing/unpausing of the stream with headphone play/pause buttons. This is a feature I found annoying to be missing when passively watching the office hours stream yesterday.

This also works in picture in picture mode (tested on android, would appreciate iOS testing)

The stream just shows a black screen when paused, it would be better to show a freeze frame of the stream when paused but I wasn't sure of how to implement this - happy to add if someone has pointers on how to do it.

### DIFF
--- a/js/components/src/components/mobile-player/use-media-button.native.tsx
+++ b/js/components/src/components/mobile-player/use-media-button.native.tsx
@@ -1,0 +1,16 @@
+import { useEffect } from "react";
+import { DeviceEventEmitter } from "react-native";
+
+export function useMediaButton(onPress: () => void) {
+  useEffect(() => {
+    const subscription = DeviceEventEmitter.addListener(
+      "mediaButtonPress",
+      () => {
+        onPress();
+      },
+    );
+    return () => {
+      subscription.remove();
+    };
+  }, [onPress]);
+}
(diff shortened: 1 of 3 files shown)


iameli:
Oooh let me think about this. There's an argument to be made that it should just mute the stream? When we have full DVR support it can then _truly_ pause the livestream but until then it'll just have skipped forward when you come back

JordanWeatherby:
> Oooh let me think about this. There's an argument to be made that it should just mute the stream? When we have full DVR support it can then _truly_ pause the livestream but until then it'll just have skipped forward when you come back

Yeah, Twitch livestreams just pause and then skip forward to present when unpaused and I find that much preferable to the play/pause button just doing...nothing. I don't hate it muting/unmuting but think that would be a little bit of unexpected UX.

Ultimately up to you, won't take it personally if you decide to just put this one off until further DVR functionality.

espeon:
FWIW twitch actually pauses the stream, which matches this PR?

JordanWeatherby:
@iameli I'd forgotten about this but it popped back up for me in my list of PRs in Arc 

DVR support seems to be more fleshed out now than it was when I wrote this

What's the move here? Happy to rebase, I do think this is a net value add, but equally happy to just close this if you have grander ideas about the direction of DVR

JordanWeatherby:
Alright well I'm gonna close this

PR #930, opened 2026-02-18
fix: misc player improvements

espeon:
- Gradient at bottom of screen instead of plain bar
- Volume bar fades out when not actively being used
- Large red unmute button works properly

image


iameli (approved):
LGTM, tested

PR #934, opened 2026-02-22
feat: audio only mode

espeon:
This PR adds a new "audio" rendition that appears in the quality menu. Users may want to listen to a livestream without the video bandwidth cost, e.g. for a podcast where the video content isn't particularly interesting.

Changed on the server:
- When the user selects the audio option in Low Latency mode (WebRTC), the server skips adding the video track over the wire.
- When the user selects the audio option in HLS playback mode, we have a new gstreamer pipeline that basically just strips the video track afaik.

Changed on the client:
- Player controls don't fade out in audio only mode
- There's a new Audio Only mode overlay
image


js/components/src/components/mobile-player/ui/audio-only-overlay.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+  content: {
+    alignItems: "center",
+    gap: 12,
+    paddingHorizontal: 24,
+  },
+  text: {
+    textAlign: "center",
+  },
+});
```
iameli: This can be made more idiomatic css per chat

iameli (approved):
Seems good, robot CSS is whatever

PR #937, opened 2026-02-24
feat: new emoji picker, with slightly less included data

espeon:


iameli (changes requested):
It should probably close after you pick one yeah? Also currently clicking the emoji picker button again doesn't close it (probably because the "click outside" and the "click button" event handlers are both running in sequence)

espeon:
Had that behaviour at first but i felt like being able to pick more than one was more important. Also will check on that

PR #939, opened 2026-02-27
Stop camera access when navigating away from go live screen

### LINKED ISSUE #920: Android app accessing camera
In the android app, hit Go Live and and then hit back to return to the home page (without going live).
The OS displays that the app is accessing your camera until the app is force quit

JordanWeatherby:
Add NativeIngestPlayer useEffect cleanup function that runs on unmount. 
This stops the OS reporting that the camera is being accessed when navigating away from the go live screen

Fixes https://github.com/streamplace/streamplace/issues/920

Tested on android but not on iOS (I assume the bug happens on iOS too but not 100% sure)


iameli:
Oh nice, I'll test and roll this into the pre-live changes that touched similar code paths

PR #943, opened 2026-03-01
feat: profile cache so we don't send hundreds of requests to the appview

espeon:


iameli (approved):
seems fine, someday we index ourselves

PR #944, opened 2026-03-01
fix mobile playback issues

iameli:
So we had a problem where chat was loading very slowly on native (in particular Android Debug, the slowest-rendering platform of them all) which was causing video-retry to trigger. The following changes have been made:

1. @JordanWeatherby I had to revert https://github.com/streamplace/streamplace/pull/921 unfortunately, something about it was breaking the memo() call and the chat lines were rerendering every few seconds - very expensive on Android
2. We now give the video up to 5 seconds to start playing before even trying to render chat; it's a slightly more awkward visual experience but gets you into a better state more quickly
3. <VideoRetry /> was being too aggressive; giving up on the first playback after 500ms was unrealistic. Bumped it up to 2s and now the initial SDP call is much more likely to succeed on mobile
4. SDP Post was completely broken on mobile whoops

### DIFF
--- a/js/app/components/mobile/ui.tsx
+++ b/js/app/components/mobile/ui.tsx
@@ -1,6 +1,7 @@
 import { useNavigation } from "@react-navigation/native";
 import {
   ContentWarningBadge,
+  PlayerStatus,
   PlayerUI,
   Slider,
   Text,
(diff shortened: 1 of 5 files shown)


JordanWeatherby:
Huh weird, sorry it caused issues

iameli:
@JordanWeatherby No worries, we do that all the time too — I reaaaaaalllly want automated UX testing with Detox to flag this kind of stuff automatically, it's been a pain in our butts 5-6 times now

PR #962, opened 2026-03-13
feat: add fullscreen 'f' + double click shortcut

espeon:


js/app/components/mobile/desktop-ui.tsx
```diff
@@ -158,17 +163,45 @@ export function DesktopUi({
...
+        toggleFullscreen();
+      }
+    }
+
+    document.addEventListener("keydown", handleKeyDown);
+
+    return () => {
+      document.removeEventListener("keydown", handleKeyDown);
+    };
+  }, [toggleFullscreen]);
```
iameli, line 190: You can still type Fs in chat and login and stuff yeah?

iameli (approved):
Looks good as long as you can still f in the chat to pay respects etc

js/app/components/mobile/desktop-ui.tsx
espeon: Ooh, good point. Seems to be fine with the chat's input but will put something in there just in case.

iameli: Yeah looks good!

PR #964, opened 2026-03-14
feat: add params to chat popout, align better without hax

espeon:


js/app/src/screens/chat-popout.tsx
```diff
@@ -43,8 +62,15 @@ export function PopoutChatInner({ user }: { user: string }) {
...
         ]}
       >
-        <Chat />
-        {profile && (
+        <Chat
+          {...(reverseChat || hideAfter
+            ? ({
+                ...(reverseChat ? { reverse: true } : {}),
+                ...(hideAfter ? { hideAfter } : {}),
+              } as any)
```
iameli: Hahahaha I can just feel your frustration with this `as any`

espeon: lmao, rethinking that and i'm pretty sure i can simplify that by a lot. thinking that probably stemmed from these LSP issues I keep on getting

iameli (approved):

PR #968, opened 2026-03-17
fix: add styling to HLS to have it not be 0x0px on mobile

espeon:


espeon:
this is kinda important so gonna go ahead and merge 🙇🏼

PR #974, opened 2026-03-21
fix: add 'collapse sidebar' buttons to every stack possible

### LINKED ISSUE #973: "Retract sidebar" button is not on some headers
title, basically. Have spotted this on most routes that aren't on the main path/inaccessible from mobile. Live dashboard, about, settings etc.

espeon:
fixes #973 
We seem to have missed a bunch of these when transitioning to the new tab-driven layout  so i've added analogues to that everywhere there wasn't one.


iameli (approved):
LGTM

PR #978, opened 2026-03-23
feat: roll of film player

espeon:


iameli:
Added to https://github.com/streamplace/streamplace/pull/1137

PR #979, opened 2026-03-23
feat: chat pinning

espeon:
## TODO: description


iameli (changes requested):
Looks great overall!

Seems like you can't pin your own messages

If we get a `/iame.li/pinned-message` style route we can put it on iPads or whatever for the conference

pkg/model/pinned_record.go
```diff
@@ -0,0 +1,95 @@
...
+	IndexedAt     *time.Time `gorm:"column:indexed_at"    json:"indexedAt"`
+	ExpiresAt     *time.Time `gorm:"column:expires_at"    json:"expiresAt"`
+	CreatedAt     time.Time  `gorm:"column:created_at"    json:"createdAt"`
+}
+
+func (p *PinnedRecord) ToStreamplacePinnedRecord() (*streamplace.ChatPinnedRecord, error) {
+	rec := &streamplace.ChatPinnedRecord{
+		LexiconTypeID: "place.stream.chat.pinnedRecord",
+		PinnedMessage: p.PinnedMessage,
+		CreatedAt:     p.CreatedAt.UTC().Format(time.RFC3339),
```
iameli: I've been using `.Format(util.ISO8601)` more lately, there's a distinction but I don't remember what it is but indigo uses the ISO one so

pkg/model/pinned_record.go
```diff
@@ -0,0 +1,95 @@
...
+	"context"
+	"errors"
+	"time"
+
+	"gorm.io/gorm"
+	"stream.place/streamplace/pkg/streamplace"
+)
+
+type PinnedRecord struct {
+	RKey          string     `gorm:"primaryKey;column:rkey"`
```
iameli: We more commonly use URI instead of RKey as the primary key

Wait actually this can't be the primary key 'cause technically different users can have the same rkey

espeon: whoops idk how this made it in 🤔

Issue #628, opened 2025-09-29
push notification: turn on or off for particular streamers


iameli:
1. Turn on or off notifications for a particular streamer
2. Opt in to "notify when anyone goes live ever"
3. Maybe even a "we think you'll like this" notification

JacobBaqleh1:
I'd like to claim this issue

PR #981, opened 2026-03-24
feat: add per-streamer notification preference

JacobBaqleh1:
Issue #628 
Open to any feedback!
## Summary
- Adds a bell/bell-off toggle to the follow button that lets followers opt out of push notifications for individual streamers
- Default behavior is unchanged — all followers still receive notifications 
- Opted-out followers are filtered before the notification blast in the queue processor

## My Test Plan
- go test ./pkg/model/...
- go test ./pkg/statedb/... 
- go test ./pkg/spxrpc/... 
- make dev — full build passes
- pnpm run check — no TypeScript errors
- GET endpoint confirms preference persists after toggle

## More Testing
Excited for more testing.  Right now, test coverage shows the opt-out logic is correct, but a full end-to-end push notification test on a real device hasn't been run yet. Also, on the web, we can add a toast notification.

## Web frontend video:
https://github.com/user-attachments/assets/1de884de-7b9b-4a45-bcfd-e47be6df9833


JacobBaqleh1:
I need to fix a privacy concern. Right now, anyone can query a user's notification preference

pkg/model/model.go
```diff
@@ -50,6 +50,10 @@ type Model interface {
 	GetUserFollowingUser(ctx context.Context, userDID, subjectDID string) (*Follow, error)
 	DeleteFollow(ctx context.Context, userDID, rev string) error
 
+	GetNotificationPreference(ctx context.Context, userDID, streamerDID string) (*NotificationPreference, error)
```
iameli: This stuff looks fine but needs to be in `stateDB`

pkg/model/notification_preference.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+
+	"gorm.io/gorm/clause"
+)
+
+type NotificationPreference struct {
+	UserDID     string `gorm:"primaryKey;index:notif_user_idx;column:user_did"`
+	StreamerDID string `gorm:"primaryKey;index:notif_streamer_idx;column:streamer_did"`
+	Enabled     bool   `gorm:"column:enabled"`
+}
```
iameli: For these types, my preference is to handle them following the pattern we set with [multistream targets](https://github.com/streamplace/streamplace/tree/next/lexicons/place/stream/multistream), which is to define an atproto `record` type, do codegen for that, and then have this struct's methods take something like `streamplace.NotificationPreference` records. This helps us keep `statedb` decoupled from the rest of the app - whoever's using this method doesn't need to know what a "statefuldb" is, it just needs to know the generated `streamplace` types.

This is also to ease the transition to _actually_ using atproto records when private data unlocks!

pkg/model/notification_preference.go
```diff
@@ -0,0 +1,49 @@
+package model
+
+import (
+	"context"
+
+	"gorm.io/gorm/clause"
+)
+
+type NotificationPreference struct {
+	UserDID     string `gorm:"primaryKey;index:notif_user_idx;column:user_did"`
```
iameli:
```suggestion
	RepoDID     string `gorm:"primaryKey;index:notif_user_idx;column:repo_did"`
```

pkg/spxrpc/notification_preference.go
```diff
@@ -0,0 +1,71 @@
...
+	if _, err := syntax.ParseDID(userDID); userDID == "" || err != nil {
+		return nil, fmt.Errorf("missing or invalid user DID")
+	}
+	session, _ := oatproxy.GetOAuthSession(ctx)
+	if session == nil {
+		return nil, echo.NewHTTPError(http.StatusUnauthorized, "authentication required")
+	}
+	if session.DID != userDID {
+		return nil, echo.NewHTTPError(http.StatusForbidden, "cannot access another user's notification preferences")
+	}
```
iameli: Could just skip this field entirely and just have you modify the setting for `session.DID`

iameli (changes requested):
Looks great! Just a few things that need to get adjusted for our conventions, in particular the model --> statedb change. tyvm!!!

js/app/components/follow-button.tsx
```diff
@@ -76,6 +79,34 @@ const FollowButton: React.FC<FollowButtonProps> = ({
...
+      return;
+    }
+    let cancelled = false;
+
+    const fetchNotificationPreference = async () => {
+      try {
+        const res = await fetch(
+          `${streamplaceUrl}/xrpc/place.stream.graph.getNotificationPreference?userDID=${encodeURIComponent(currentUserDID)}&streamerDID=${encodeURIComponent(streamerDID)}`,
+          { credentials: "include" },
+        );
```
iameli: You'll need to get a agent with `useStreamplaceAgent()` (I think) to get an authenticated call

js/app/components/follow-button.tsx
```diff
@@ -107,6 +138,29 @@ const FollowButton: React.FC<FollowButtonProps> = ({
...
+          method: "POST",
+          credentials: "include",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({
+            userDID: currentUserDID,
+            streamerDID,
+            enabled: next,
+          }),
+        },
+      );
```
iameli: Same here

JacobBaqleh1:
Hey @iameli , I updated the PR. Open to more feedback

PR #1009, opened 2026-04-01
feat: have actual a tags in chat links/handles

### LINKED ISSUE #986: [BUG] Timestamp label beside messages in stream chat wraps
At 16:4X, the timestamp broke for Streamyplace on the ATmosphereConf stream. The timestamp needs additional area to ensure it does not wrap. The best way to test this afterwards would be to use 00:00 and see if there is any wrapping. It's really strange though, because the font does appear to be monospaced, but it is unfortunately an issue.

**OS:** CachyOS (Arch Linux)
**Browser:** Zen

Image

### LINKED ISSUE #987: [BUG] Middle-clicking links does not open them in a new tab without switching away from a stream
While clicking on a link does *not* replace the tab where you are watching a stream, and creates a new tab, it is not possible to middle-click to avoid being automatically redirected to the new tab that was created. This appears to be a bug.

**OS:** CachyOS (Arch Linux)
**Browser:** Zen

espeon:
(also add them back to sidebar)
Screenshot 2026-03-31 at 5 23 07 PM

### DIFF
--- a/js/app/components/sidebar/sidebar-item.tsx
+++ b/js/app/components/sidebar/sidebar-item.tsx
@@ -90,7 +90,7 @@ export default function SidebarItem({
       role="link"
       accessibilityLabel={typeof label === "string" ? label : "Link to " + href}
       // @ts-ignore This makes it render as <a> on web!
-      href={route ? href : undefined}
+      href={href}
     >
       <View
         style={[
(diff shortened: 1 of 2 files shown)


espeon:
included in #1030

PR #1010, opened 2026-04-01
fix: fill in space with chat

espeon:
When the video rises b/c the top has disappeared, the chat should fill in the remaining space. also video shouldn't rise on desktop lol


espeon:
included in #1030

PR #1011, opened 2026-04-01
fix: set an actual chatbox width when popping out

### LINKED ISSUE #988: [BUG] Popping out stream chat does not set the size of the new window, leading to identical size to original window
Title is self-explanatory, pop-out stream chat is not the correct size, so it ends up being absurdly large, especially if you've opened it on an ultra-wide monitor. This should be resized to the maximum size that the stream chat is supposed to be, which I assume it 480px based on the size that it is within Streamyplace.

**OS:** CachyOS (Arch Linux)
**Browser:** Zen

Image

espeon:


felinusfish:
Don't forget to write "fixes #988" in the description so the issue is linked and automatically closes when merged

edit: oh i see, you've been manually doing it. lol. well there's a tip for you

espeon:
included in #1030

PR #1016, opened 2026-04-01
feat: logarithmic scale for slider

### LINKED ISSUE #994: [BUG] Volume is linear instead of logarithmic, meaning perceived volume is not accurately reflected based on the slider
Ideally, the volume slider should be logarithmic, so that the slider accomplishes perceived volume, rather than +1 and -1 to volume decibels.

espeon:
closes #994 

it's technically a square/sqrt but i think this provides the most reasonable option vs like a cube root or w/e


felinusfish:
pog!! +1 to this. should improve volume control for people on desktop :)

espeon:
included in #1030

PR #1018, opened 2026-04-02
fix: persist profilecard

### LINKED ISSUE #993: [BUG] Profile pop-up modals close if a new message appears in stream chat
The profile pop-up modal closes if a new message appears while you are looking at one during a stream. This makes it impossible to read profiles in fast-moving chats.

espeon:
Closes #993


espeon:
included in #1030

PR #1020, opened 2026-04-02
feat: hide chat on popout

### LINKED ISSUE #990: [BUG] Sidebar chat should be hidden once popped out to reduce unnecessary network usage
Relatively self-explanatory, just hiding the sidebar chat once it has been popped out, similar to Twitch. This would be to reduce unnecessary network usage, since having two chats at the same time isn't needed.

**OS:** CachyOS (Arch Linux)
**Browser:** Zen

Image

espeon:
Should fix #990, unsure about the prop passing (we can maybe have this as a player control or something?)


espeon:
included in #1030

PR #1022, opened 2026-04-02
feat: slight profile redesign

### LINKED ISSUE #992: [BUG] Usernames mentioned redirect to Bsky, profiles however do not
When you click on a username mentioned in the stream chat, you are redirected. This does not happen for the profiles when clicking on a username in the chat from somebody that sent a message instead. Profiles also close after a new message appears in the chat, rather than staying open.

Image

espeon:
More compact design, with a link to the user's bluesky profile
| before | after |
|-|-|
|image|image|


iameli (approved):

espeon:
included in #1030

PR #1024, opened 2026-04-03
fix: restrict errors when making initial profiles by a lot?

espeon:
Thinking we shouldn't make profiles when we can't reach the PDS? Seems to work okay but may need to test again when exposing to public internet


espeon:
included in #1030

PR #1028, opened 2026-04-04
fix: scroll up and down on chat pickers

### LINKED ISSUE #1000: [BUG] Pressing up and down on the auto-complete list can cause the selection to disappear off-screen
The selection can go off-screen and does not scroll along with the list, making it impossible to select an emoji that goes under the emoji selection.

Image

espeon:
Web only for now, but this was very requested + annoying me also

https://github.com/user-attachments/assets/bf38f199-c858-48e9-9290-6728b5072555


espeon:
included in #1030

espeon:
nm

iameli (approved):

PR #1033, opened 2026-04-09
add mobile app banner on non-safari ios + android

espeon:
image

### DIFF
--- a/js/app/components/mobile-app-banner.tsx
+++ b/js/app/components/mobile-app-banner.tsx
@@ -1,0 +1,103 @@
+import { Text, View } from "@streamplace/components";
+import {
+  bg,
+  borders,
+  colors,
+  flex,
+  gap,
+  layout,
+  px,
+  py,
+  r,
+} from "@streamplace/components/src/lib/theme/atoms";
+import { Image } from "expo-image";
+import { X } from "lucide-react-native";
+import { useState } from "react";
+import { Linking, Platform, Pressable } from "react-native";
+import { useSafeAreaInsets } from "react-native-safe-area-context";
+
+const DISMISSED_KEY = "mobile_app_banner_dismissed";
+
+function getMobilePlatform(): "ios" | "android" | null {
+  if (Platform.OS !== "web") return null;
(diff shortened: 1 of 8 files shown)


js/app/components/mobile-app-banner.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const ua = navigator.userAgent;
+  const isIOS = /iPhone|iPad|iPod/.test(ua);
+  const isSafari = /Safari/.test(ua) && !/CriOS|FxiOS|EdgiOS/.test(ua);
+  const isAndroid = /Android/.test(ua);
+  if (isAndroid) return "android";
+  // iOS Safari shows the native Smart App Banner via the apple-itunes-app meta
+  // tag, so only show our custom banner for other browsers on iOS.
+  if (isIOS && !isSafari) return "ios";
+  return null;
+}
```
iameli: We've got some logic for this already, you might need to expand slightly but it should be usable https://github.com/streamplace/streamplace/blob/d6ba2c8fb1127a0bd6ea1c7bfdec992557bd7b7b/js/app/hooks/usePlatform.tsx#L16-L34

js/app/components/mobile-app-banner.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const STORE_URLS = {
+  ios: "https://apps.apple.com/us/app/streamplace/id6535653195",
+  android: "https://play.google.com/store/apps/details?id=tv.aquareum",
+};
```
iameli: Let's put these in some shared location and then they can be referenced here as well https://github.com/streamplace/streamplace/blob/d6ba2c8fb1127a0bd6ea1c7bfdec992557bd7b7b/js/app/components/get-apps.tsx#L8-L37

iameli (changes requested):

js/app/components/mobile-app-banner.tsx
espeon: Good point, totally missed that + the above

PR #1041, opened 2026-04-15
feat: install wasm32-wasip1 target when building muxl-wasm

espeon:
totally missed that this wasn't installed when building 🤨 so now i've added it so nobody else ever misses it again 🫨


iameli:
sure!

PR #1048, opened 2026-04-17
fix: Clear reply on escape in chat box

espeon:


iameli (changes requested):
syntax errorrrrrrrrrrrr

iameli (approved):

PR #1049, opened 2026-04-17
feat: add author handle resolution when repo unavailable

espeon:


pkg/atproto/atproto.go
```diff
@@ -182,6 +186,25 @@ func (atsync *ATProtoSynchronizer) RefreshIdentity(ctx context.Context, did stri
...
+	if err != nil {
+		log.Warn(ctx, "failed to resolve author handle", "did", author.Did, "err", err)
+		return
+	}
+	if repo != nil && repo.Handle != "" {
+		author.Handle = repo.Handle
+		handleCache.SetDefault(author.Did, repo.Handle)
+	}
+}
+
```
iameli, line 207: don't love that this mutates (and also like you said should prolly just do handle resolution instead of sync)

iameli (changes requested):

PR #1050, opened 2026-04-22
feat: stream topics and tags

### LINKED ISSUE #588: Add tags to streams (e.g. language/game tags)
Right now streams can only be described by their title, and soon by metadata based on what is physically happening in the video by c2pa tags. It would be, then, super useful to allow attaching out-of-band tags to streams. This can take the form of game tags (`Chess`, `League of Legends`, `Minecraft`) based probably on IGDB, or other prior agreed-upon tags (e.g. `speedrun`, `chatting`, `music production`). C2PA should take care of language tags.

This helps others discover streams more easily by filtering on interests, languages, and other categories, it enables better filtering once streamplace gets larger, and provides a more consistent system for the previous other than relying on text-only titles.

In my opinion, the best way to do this would be some sort of tags field, and a 'activity' field. Activities can include games, and other non-gaming activities, and tags would include the rest. Activity should be `string`, but tags should be `string[]`. 

We'd want to:
- [ ] display these on the stream page and in various listings, e.g. the home page.
- [ ] Add autocomplete for activities and other tags

In the future, it would be nice to:
- [ ] filter by these listings in some sort of search page 
- [ ] have user-defined tags, or a way for users to suggest tags
   - I don't think users should be able to put arbitrary tags (even though it would be allowed by the lexicon, we can choose not to show them in the appview)
- [ ] have protocol-wide trending tags?

espeon:
Closes #588


iameli (changes requested):
Minor bugs:

* Adding a tag makes you lose focus on the tag-adding box, making it annoying to add a bunch
* I added "rocketleague" and "scumbag" as tags but only one seems to show up here:

Image

PR #1063, opened 2026-04-27
feat: readd active tab highlighting to sidebar

espeon:
this basically. we used to have it to give better signalling to where we are but it somehow got lost in the last big refactor
image


iameli (approved):
So many special cases, ` if (target.tab === "SettingsTab") return true;` and whatnot

I'm cool with this but it'd be nice to identify the root causes why our model is lining up with react-navigation so inconsistently

espeon:
It's primarily because our sidebar is manually populated and not actually touched at all by react navigation. I'll do a look over and see if I can simplify some of that though?

PR #1064, opened 2026-04-27
feat: persist badge and chat state management

espeon:
persist chat to fix https://bsky.app/profile/aka.dad/post/3mkizdg7t3k2i


iameli (approved):
Untested but seems broadly reasonable

PR #1065, opened 2026-04-27
fix: can not go to accounts page from certain views

espeon:
we probably should not be using hooks directly in sideeffects


iameli (approved):
oh yeah that seems sketchy

PR #1066, opened 2026-04-28
fix: move badge switcher to profile

espeon:
it was originally designed to be in profile so that is where it will go


iameli (commented):
Seeing lots of new translation strings but not a commit with the actual translations?

PR #1073, opened 2026-05-05
fix: only have a single theme context

espeon:
Our theme stuff includes a portal host and in certain situations we had multiple portal hosts in our theme context, which led to duplicate items when they are portaled


iameli (approved):
ohhhhhhhhhhhhhhh

PR #1112, opened 2026-05-18
feat: bio

espeon:
Adds `place.stream.bio.*`, a new lexicon namespace for customizable creator profile pages. Also, a full bio viewer, settings UI, and a import path from Leaflet public pages. Also hooked up moderation.

Example: 
image


iameli (changes requested):
Neat! How do I get to different columns and stuff? I see there's Leaflet import but I feel like if we don't ship an editor for most of the stuff yet we shouldn't ship the lexicons until we're ready

This link seemed to not work:

Image

espeon:
Columns should be in the column editor when you import a leaflet doc? I think when we make the actual editor we'll have some other way of specifying columns (like having the editor actually have a column primitive).

Editors in react native usually are pretty bad but I guess I can try my hand at making one haha

espeon:
hmmmmmm definitely will have to rebase this

PR #1121, opened 2026-05-23
feat: rows on mobile!

espeon:
we've been growing enough to warrant a more space-efficient way of featuring people on mobile! this does that, including a bunch of small tweaks to make home look better on those devices. 

will say on mobile it's kind of crammed in there, unsure if it's worth messing with, every alternative would change the design pretty drastically IMO.

Screen Shot 2026-05-22 at 20 35 15


iameli:
Nice! Should test on iOS/Android and I'll merge

iameli (changes requested):
lil stretched weird on iOS per stream

PR #1129, opened 2026-05-27
media: auto transcode Opus/AAC in all cases

iameli:


iameli:
incorporated in https://github.com/streamplace/streamplace/pull/1131

PR #1139, opened 2026-05-31
fix: delegate and persist pins

espeon:
Adds option to delegate pins to moderators, plus have previous pins in the repo (but still only display the last one).


timconspicuous:
Thank you :)

iameli (approved):

PR #1171, opened 2026-06-25
atproto: consume moqt:// (MoQ-over-QUIC) relays in the firehose

iameli:
--relay-host entries may now be moqt:// (moq/moql/moqs) URLs alongside ws://wss://, consumed over QUIC via github.com/streamplace/atmoq-go. A MoQ frame is byte-identical to a subscribeRepos WebSocket message, so it decodes through the same indigo event types and feeds the same scheduler, dedup, cursor, and backoff machinery as the WebSocket path — a relay just appears in the comma-separated list and cross-relay-dedups against the others.

connectRelay branches on URL scheme; the shared event callbacks are factored into repoStreamCallbacks. connectRelayMoq dials/subscribes and pumps frames through dispatchMoqFrame (mirrors HandleRepoStream's decode switch, routing #commit/#identity). MoQ is live-edge only, so the relay cursor is observed for liveness but not used to resume.


iameli:
replaced by https://github.com/streamplace/streamplace/pull/1172

PR #1179, opened 2026-07-07
Eli/idle finalize guard

iameli:
## Summary by CodeRabbit

* **New Features**
  * Added “Podcasting” as a supported activity label across the app and reference docs.

* **Bug Fixes**
  * Improved livestream transitions so starting a new livestream best-effort closes any prior one for the same repo.
  * Prevented premature livestream finalization by rescheduling when a livestream is the latest but still stale.

* **Tests**
  * Added coverage for livestream rescheduling and for ensuring superseded livestreams aren’t handled via the reschedule path.

### DIFF
--- a/js/components/src/lib/metadata-constants.ts
+++ b/js/components/src/lib/metadata-constants.ts
@@ -3,6 +3,7 @@ import { PlaceStreamDefs, schemas } from "streamplace";
 // "knownValues": [
 //   "events",
 //   "just_chatting",
+//   "podcasting",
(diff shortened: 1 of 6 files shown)


iameli (commented):
I don't love this implementation; it seems like it makes the queue really complicated and I want to avoid global locking in statedb as much as possible because it's the kind of thing that can bring Streamplace to a screeching, grinding halt. But this bug just bit me, so fuck it for now.

PR #1180, opened 2026-07-07
js lexicons: migrate to @atproto/lex

iameli:
This moves us from @atproto/lex-gen to @atproto/lex, ignore coderabbit lmao


iameli:
nah we're doing both at once lfg

PR #1187, opened 2026-07-10
Add stream.received webhook event

mackswan:
**Summary**

Adds a new stream.received webhook event that is enqueued once per StreamSession when the first local media segment is accepted.

This is intended to let integrations react when media has actually started arriving, without waiting for a published livestream record.

Changes included:

Add stream.received to webhook lexicon event enums.
Expose stream.received in the webhook settings UI.
Enqueue a stream_received task after a local segment is accepted.
Send the event through the existing webhook manager / Discord-shaped webhook sender path.
Add a statedb test for looking up active stream.received webhooks.
**Notes**

The event is enqueued once per StreamSession using a session-local guard. Delivery follows the existing webhook task behavior.

**Testing**

Focused package tests passed:

PKG_CONFIG_PATH=/home/ecash/dev/streamplace/build-linux-amd64/lib/pkgconfig:/home/ecash/dev/streamplace/build-linux-amd64/lib/gstreamer-1.0/pkgconfig:/home/ecash/dev/streamplace/build-linux-amd64/meson-uninstalled LD_LIBRARY_PATH=/home/ecash/dev/streamplace/build-linux-amd64/lib CGO_LDFLAGS="-lm" go test ./pkg/director ./pkg/statedb ./pkg/integrations/webhook ./pkg/integrations/discord

Result:

ok stream.place/streamplace/pkg/director
ok stream.place/streamplace/pkg/statedb
? stream.place/streamplace/pkg/integrations/webhook [no test files]
? stream.place/streamplace/pkg/integrations/discord [no test files]

Full make dev-test was attempted locally but blocked by an existing local GStreamer runtime issue in pkg/multitest:

failed to create appsrc element: could not create element: appsrc

### DIFF
--- a/pkg/integrations/discord/send-stream-received.go
+++ b/pkg/integrations/discord/send-stream-received.go
@@ -1,0 +1,52 @@
+package discord
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"stream.place/streamplace/pkg/aqhttp"
+	"stream.place/streamplace/pkg/integrations/discord/discordtypes"
+)
+
+func SendStreamReceived(ctx context.Context, w *discordtypes.Webhook, streamerDID string) error {
+	content := fmt.Sprintf("stream.received %s", streamerDID)
+	for _, rewrite := range w.Rewrite {
+		content = strings.ReplaceAll(content, rewrite.From, rewrite.To)
+	}
+
+	payload := discordtypes.Payload{
+		Content: fmt.Sprintf("%s%s%s", w.Prefix, content, w.Suffix),
+	}
+
+	jsonPayload, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("failed to marshal payload: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", w.URL, bytes.NewReader(jsonPayload))
+	if err != nil {
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := aqhttp.Do(ctx, req)
+	if err != nil {
+		return fmt.Errorf("failed to send request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusNoContent {
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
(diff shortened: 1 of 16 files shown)


pkg/integrations/discord/send-stream-received.go
```diff
@@ -0,0 +1,52 @@
...
+	"io"
+	"net/http"
+	"strings"
+
+	"stream.place/streamplace/pkg/aqhttp"
+	"stream.place/streamplace/pkg/integrations/discord/discordtypes"
+)
+
+func SendStreamReceived(ctx context.Context, w *discordtypes.Webhook, streamerDID string) error {
+	content := fmt.Sprintf("stream.received %s", streamerDID)
```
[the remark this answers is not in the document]
WonkWorks: Thanks — leaving this minimal for the first version so the event follows the existing content-based webhook path without adding profile/PDS resolution to the enqueue/dispatch flow. Happy to enrich the payload in a follow-up if that is the preferred direction.

pkg/statedb/queue_processor.go
```diff
@@ -483,6 +491,43 @@ func (state *StatefulDB) processNotificationTask(ctx context.Context, task *AppT
...
+func (state *StatefulDB) processStreamReceivedTask(ctx context.Context, task *AppTask) error {
+	var streamReceivedTask StreamReceivedTask
+	if err := json.Unmarshal(task.Payload, &streamReceivedTask); err != nil {
+		return err
+	}
+
+	webhooks, err := state.GetActiveWebhooksForUser(streamReceivedTask.StreamerDID, "stream.received")
+	if err != nil {
+		log.Error(ctx, "failed to get stream.received webhooks", "err", err)
+	} else {
```
iameli: I think for a lot of these cases we'll want to actually return an error (to allow for consistent logging and retry in the future. Also cleaner if you return early, you can drop the `if` block, like:


```suggestion
	if err != nil {
		return fmt.Errorf("failed to get stream.received webhooks: %w", err)
	}
```

iameli: Fair enough that you were following the other functions pretty much exactly though, lol

iameli (approved):
I made one suggestion and then saw that you were perfectly matching the existing code, lol

Thanks for the contribution! Will merge when CI passes

WonkWorks:
Thanks again for merging this. No rush, but do you know whether
stream.place deploys the web app from next automatically, or will
stream.received show up in the webhook UI only after the next
release/deploy?

PR #1188, opened 2026-07-10
dev: make plain 'go test' work on Linux with no env vars

iameli:
Three pieces:

- hack/pkg-config.sh: a pkg-config wrapper that injects the meson build dir's pkgconfig paths, but only for modules our build provides (streamplacedeps, glib/gstreamer families, lpms's libav*). On Linux --libs answers also gain -lm and an old-style rpath to the build dir, so ld resolves transitive .so deps at link time and test binaries run without LD_LIBRARY_PATH. Explicit PKG_CONFIG_PATH still wins, and everything else passes through untouched.

- make dev-setup now registers the wrapper via 'go env -w PKG_CONFIG=...', which Go persists in its own per-user config -- no shell env needed, and gopls picks it up too.

- streamplacedeps.pc gains -Wl,-rpath,${libdir} -Wl,--disable-new-dtags on Linux shared builds; DT_RPATH (unlike DT_RUNPATH) also covers transitive loads like libgobject -> libffi and libavcodec -> libswresample, whose meson-built libraries carry no rpath.

Note: a build dir produced inside a container (prefix=/app) bakes wrong paths into every .pc file; rerun 'make dev-setup' natively after using one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,6 +292,13 @@ test-vod:
 .PHONY: dev-setup
 dev-setup:
 	$(MAKE) -j16 app-cached dev-setup-meson
+	$(MAKE) dev-go-env
+
+# Point Go's cgo at our pkg-config wrapper (persisted in `go env`, no shell
+# env vars needed) so plain `go test ./pkg/...` finds the meson-built deps.
+.PHONY: dev-go-env
+dev-go-env:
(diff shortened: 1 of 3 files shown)


iameli:
Closed in favor of https://github.com/streamplace/streamplace/pull/1190

PR #1192, opened 2026-07-13
Fix webhook event label rendering

mackswan:
## Summary

Fixes a frontend crash in the webhook settings page when rendering saved webhook event labels for event values containing dots, such as `stream.received`.

The backend event value is unchanged. This only updates the UI label lookup to use the existing event option label keys instead of deriving an i18n key directly from the raw event value.

## Testing

- Ran `pnpm run check` from `js/app`

### DIFF
--- a/js/app/components/settings/webhook-manager.tsx
+++ b/js/app/components/settings/webhook-manager.tsx
@@ -96,6 +96,15 @@ const EVENT_OPTIONS = [
   { value: "stream.received", labelKey: "events-stream-received" },
 ];
 
+const EVENT_LABEL_KEYS = new Map(
+  EVENT_OPTIONS.map((option) => [option.value, option.labelKey]),
+);
+
+function getEventLabel(t: (key: string) => string, event: string) {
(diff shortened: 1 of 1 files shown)


iameli (approved):

WonkWorks:
Hey Eli — I think I found a backend issue with webhook listing.

I can consistently reproduce the following on stream.place:

Create a webhook with Active = true → it saves successfully but does not appear in the Webhooks list.
Create the same webhook with Active = false → it appears immediately in the list.
Edit that webhook and set Active = true → it disappears from the list again.

The interesting part is that the webhook itself still appears to function. Once I corrected my GAS endpoint (?source=streamplace) it successfully received stream.received events and posted to Bluesky, so this seems to be a listing/UI/backend issue rather than webhook delivery.

I dug through the current source with Codex. The checked-in code doesn't seem to explain this behavior—the frontend isn't filtering on active, and listWebhooks should return all webhooks when no active query parameter is supplied. My HAR also shows the request is just GET /xrpc/place.stream.server.listWebhooks?limit=50 with no active filter.

Happy to send the HAR if it's useful.

PR #1203, opened 2026-07-19
bus: do not buffer dead segments

espeon:
Pls merge after #1204


espeon:
think i added this fix in a better way in another PR

PR #1205, opened 2026-07-20
fix: WebRTC frame loss at keyframes

espeon:
Noting that I don't seem to hit this path after I reset OBS settings, so would treat this as more or less untested.

Frame loss at every keyframe on certain streams, WebRTC (WHEP) playback only (HLS and multistream-RTMP outputs unaffected). Triggered I think by a stream ending up in a duplicated/multitrack ingest state by OBS.


espeon:
will split this actually

PR #1206, opened 2026-07-21
media: pin h264timestamper to no reordering via a vendored property

espeon:
add a max-reorder-frames parameter so we don't do extremely bad things (delay by a guessed, yet almost always incorrect framerate) if the SPS lacks VUI params


iameli:
Should be worked around by https://github.com/streamplace/streamplace/pull/1210 but we can revisit this if that causes other problems

PR #1207, opened 2026-07-21
fix segment ordering

espeon:


espeon:
nm this was already implemented

PR #1235, opened 2026-08-02
feat: add "stream duration" pin length

espeon:


lexicons/place/stream/chat/pinnedRecord.json
```diff
@@ -29,6 +29,16 @@
             "type": "string",
             "format": "datetime",
             "description": "Optional expiration time. If set, the pin is considered inactive after this time."
+          },
+          "duration": {
+            "type": "string",
+            "knownValues": ["streamEnd", "forever"],
+            "description": "How long this pin lasts when expiresAt is not set. 'streamEnd' (the default when unset) keeps the pin active until the livestream referenced by 'livestream' ends. 'forever' keeps the pin active across streams until manually unpinned."
+          },
+          "livestream": {
```
iameli, line 33: Couldn't the semantics just be that empty livestream == forever? Do we need a new field?

iameli (changes requested):
i think maybe we don't need duration

lexicons/place/stream/chat/pinnedRecord.json
espeon: fixed (a month ago)

PR #1261, opened 2026-08-18
Lift up stream settings so the metadata doesn't revert on rerender

espeon:
When crossing the 1200px boundary, we totally reset the live dashboard. So, to rerender the components with the same data, we lift up the stream settings outside of the component and put it in a context.

Ideally we would reflow, but this technique can give us slightly more control if we use the components library in other apps.


espeon:
we should just remake this page, probably not worth this effort if so

PR #1272, opened 2026-09-03
hls: low latency mode

espeon:
Adds LL-HLS playback alongside regular HLS at endpoint `/api/playback/<user>/llhls/main.m3u8`. Verified playback on MSE and Safari.

Note that I updated `gst-plugins-rs` here as a patch for testing, would need that updated in our gstreamer fork


espeon:
probably will not merge as it will be more or less unneeded if we end up using moq

PR #1275, opened 2026-09-11
Preserve custom facet features in websocket messages

espeon:
## Summary

- Preserve and decode unknown rich-text facet variants when serializing websocket chat messages.
- Add relevant tests


espeon:
whoops ill delete

PR #1296, opened 2026-09-22
agent field guide

iameli:


AGENTS.md
espeon: I feel like some of this is useful operational knowledge, but I would probably not accept this as a canonical repo-wide AGENTS.md or other shared documentation file. 

The document mixes otherwise good info about the repository with assumptions about your specific developer environment, which almost certainly no one else shares. For example, agents and others who read this document are told they are working under `~/code/streamplace-N`, assumes sibling checkouts, fish, specific container names, and later hard-codes `/home/iameli/code/...` into docker, build, and other instructions.

As AGENTS.md is generally consumed as instructions and requirements, an agent who reads this file working from a machine that isn't configured with your exact setup will execute instructions that simply don't apply, which could, in the worst case, have disastrous effects. There's also a lot of pretty important machine level instructions here (e.g. starting and stopping docker containers, killing processes, and messing with local git behaviour) which one may not be so comfortable with agents running on a machine with other personal/work data.

I'm also worried about the readability of this document; AGENTS.md et al can be [very useful to human contributors who may not want to use LLMs](https://bsky.app/profile/did:plc:x5leeami62gbmjekf7ro6sqo/post/3mvxw77ifoc2k), and this is dense enough to create too much friction, affecting parseability and usefulness. I think the same problem applies to agents. Tepository invariants, machine-specific setup, troubleshooting history, known failures, and workflow conventions are all presented with the same weight, making it difficult to tell which instructions are actually universal, and which ones apply to your development environment

I'd recommend splitting this into:
- A shorter, more conventional AGENTS.md with environment independent instructions, such as build/test entry points, file rules, some gotchas, CI expectations and PR conventions.
- A document associated with the specific 'provisioned' setup elsewhere, possibly mentioned from the AGENTS.md.

espeon: Hmm, looking at the rest of the PR, I kind of get where you're going but I still think promoting one specific agentic workflow into the repository's canonical development workflow, and I'm not sure that's a good idea.

People tend to work fairly differently. Some prefer working one checkout at a time, some prefer worktrees, and others may not want to use agents at all. I don't think the repo-wide defaults should assume one specific way of orgaanizing unless we're deliberately deciding this is now the current happy path development model. 

The repo, I think, should instead expose stable primitives (for example, as build/test/provision/e2e skills) without prescribing nearly this much about how someone's overall dev env is built around them.

